PR #6, opened 2016-09-08
MAINT: clean up some outdated/irrelevant links.

rgommers:
No need for that to be on the front page. 

We should add some new content soon, like link to governance docs and NumFOCUS.


jaimefrio:
Thanks Ralf!

rgommers:
now finally uploaded....

PR #11, opened 2018-03-08
Remove www from CNAME

stefanv:


rgommers:
Maybe add a short description/rationale. Did the current `CNAME` cause an issue somewhere?

stefanv:
The idea was to make the URLs shorter and more modern.  But it turns out there are some subtleties in moving from `www.numpy.org` to `numpy.org` (https://help.github.com/articles/about-supported-custom-domains/#www-subdomains), so I'm going to close for now.

PR #13, opened 2018-09-02
Add CoC link in sidebar.

rgommers:
Also move stuff around to avoid warnings. Using `www/` is the same structure as for https://github.com/scipy/scipy.org


rgommers:
No comments, clear improvement - merging

PR #19, opened 2019-04-10
ENH: donate via salsalabs

mattip:
Change button to salsalabs


mattip:
I checked that the link works but did not try to donate :)

rgommers:
I tried, all works as expected:)

rgommers:
thanks Matti

PR #21, opened 2019-05-17
Add NumFOCUS, sponsor, and institutional partner logos

rgommers:
This is another band-aid to acknowledge sponsorship and institutional partners. It's not super pretty but it works. We need a Tidelift reference on the site. I adapted the ordering and text from Jupyter (https://jupyter.org/about).


rgommers:
I think this is good to merge. A request for help though: if you're reviewing a html/css thing that should be done better, please feel free to push to my branch. I'm not a very talented web dev.....

seberg:
I really do not want to delay this, but I am slightly worried about adding it before having an intuition on where to draw the line for Institutional partners/sponsors. Which is OK, it does not have to be me who has that intuition ;).

Jupyter seems to also list the Slate and Moore foundations, should do that as well @stefanv?

I believe scikit-learn also has a lot of experience with doing this right (just in case we want to get an external opinion), although they are a much more complex project in this regard.

rgommers:
> Jupyter seems to also list the Slate and Moore foundations, should do that as well @stefanv?

Yes, good point, definitely.

rgommers:
> I really do not want to delay this, but I am slightly worried about adding it before having an intuition on where to draw the line for Institutional partners/sponsors

Yeah, it's worth spelling out a few rules for this. Institutional Partner rules are clearly formulated in our governance doc, but for sponsors that is not the case. I'd suggest:
- It must be active, or if a one-off donation, then within the last 12 months.
- There must be some minimum level. Perhaps $10k/yr?

We can always revise this of course.

seberg:
Sure, I would personally prefer time approximations over money (in case it has to do with developer time specifically), but that makes little difference.

Even for institutional partners there is some potential for conflict. If someone working for company XY is asked to fix a bug in numpy as their official job duty, that seems not enough to get free advertisement for XY, it would have to be substantial and continued effort, I guess. I did not bring that up earlier, because for me there is the step of disclosure for possible conflict of interest and then that of "advertising" in a sense. Anyway, just my quick thoughts...

rgommers:
> Sure, I would personally prefer time approximations 

We had that discussion for institutional partners, and decided on purpose to not put a time amount. It is clear that 30 min/wk is not enough though.

For money it's much easier though, and we do need money not just dev time. Given that this is basically valuable "advertising space", a number would be good to have. $10k is in line with the lowest sponsorship level at NumFOCUS.

rgommers:
oh and, dev time > money, that's the rationale for putting Institutional Sponsors higher up the page.

rgommers:
Added the Sloan and Moore logos.

I'll write to the mailing list to discuss some guidelines for this. That may take quite a while to do those (weeks at least), so it should not stop merging this as is I think.

rgommers:
Sent proposal for guidelines to the mailing list. I'd like to get this in though, it's already well overdue (Tidelift started paying us for April already). **I will merge this tomorrow unless there are serious objections.** we can adjust at any time after that

rgommers:
Okay, in it goes. Thanks for the feedback @seberg

PR #23, opened 2019-07-06
script to obtain eht-imaging dependency graph using pipdeptree and gr…

shaloo:
…aphviz and highlight numpy role in M87 imaging

Screenshot 2019-07-06 at 8 01 24 PM


mattip:
welcome @shaloo (@shaloo is an applicant to the Google Season of Documents project, with the goal of improving our documentation).

Thanks for this, I never knew about pipdeptree and this is a nice view of how NumPy is at the base of many packages.

It is not clear to me how this will be worked into the NumPy documentation. Perhaps it should be part of [`numpy/numpy.org`](https://github.com/numpy/numpy.org) instead?

rgommers:
Yes indeed, this is the right repo. It was my suggestion to create a `scripts` dir, we can integrate it with the site content later on.

rgommers:
Very nice, thanks @shaloo. This figure tells a nice story:)

This script is a little dangerous, sourcing it when you have Python installed with Conda will let `pip` install `numpy` etc. from PyPI which has the potential to mess up the whole conda env. Can you please split the script in two, one for setting up a virtualenv, and one for creating the graph? That way it's explicit that the install is being done, and people can look at it more carefully and decide to install the dependencies some other way.

shaloo:
Thanks Matti and Ralf.  Splitting it as suggested...

shaloo:
@rgommers , @mattip  Updated as suggested - pls review and merge if ok. Thx.

rgommers:
Looks good to me, merged. Thanks again @shaloo!

PR #24, opened 2019-07-06
ENH: update, don't reinitialize the numpy/numpy.github.com repo

mattip:
xref numpy/numpy#13921

Layer this top-level content on top of the existing content, so that we can keep all the content in one repo. Merging this and 13921 means we can deactivate the github pages in numpy/devdocs and numpy/neps.

Since we do not clean out the repo before adding all the generated content, files that are no longer needed will need to be deleted by hand from numpy/numpy.github.com

### DIFF
--- a/www/Makefile
+++ b/www/Makefile
@@ -18,17 +18,15 @@ help:
 
 clean:
 	-rm -rf _build/*
+	-rm -rf built_doc
 
 upload: html
-	cd _build/html && \
-	    touch .nojekyll && \
-	    echo www.numpy.org > CNAME && \
-	    rm -rf .git && \
-	    git init && \
-	    git remote add target git@github.com:numpy/numpy.github.com.git && \
-	    git add -A && \
-	    git commit -m "Rebuild site" -a && \
-	    git push -f target master:master
+	git clone git@github.com:numpy/numpy.github.com built_doc
+	cp -r _build/html/* built_doc
+	git -C built_doc add -A
+	git -C built_doc commit -m "Rebuild site" -a
+	git -C built_doc push origin
+	rm -rf built_doc
 
 html:
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) _build/html
--- a/www/index.rst
+++ b/www/index.rst
@@ -47,14 +47,23 @@ Documentation
 -------------
 
 The most up-to-date NumPy documentation can be found at
-`Latest (development) version <https://www.numpy.org/devdocs>`__.
-It includes a user guide, full reference documentation, a developer guide, meta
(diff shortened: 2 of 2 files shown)


www/Makefile
```diff
@@ -20,15 +20,12 @@ clean:
...
-	cd _build/html && \
-	    touch .nojekyll && \
-	    echo www.numpy.org > CNAME && \
-	    rm -rf .git && \
-	    git init && \
-	    git remote add target git@github.com:numpy/numpy.github.com.git && \
-	    git add -A && \
-	    git commit -m "Rebuild site" -a && \
-	    git push -f target master:master
+	git clone git@github.com:numpy/numpy.github.com built_doc
```
rgommers: This is going to be extremely slow right, if `numpy.github.com` grows really large? I'm not sure I understand the logic of recloning the generated docs here. PR title sounds useful, but don't quite see how this change achieves it.

rgommers:
> Layer this top-level content on top of the existing content, so that we can keep all the content in one repo.

The right setup is probably:
1. A numpy.org repo and deployment
2. One deployment repo for docs.

(2) gets populated from the main numpy repo as well. Also, it's really large. 

(1) is very small and may move from Sphinx to Jekyll (or similar other tool).

> Merging this and 13921 means we can deactivate the github pages in numpy/devdocs and numpy/neps.

This sounds good.

www/Makefile
mattip: I don't see how to resolve the conflict between the desire to have [fewer repos for the documentation](https://github.com/numpy/numpy/pull/13921#issuecomment-508881361) and on the other hand not clone it.

rgommers: See my other comment on this PR: one for docs, one for numpy.org

Also, it would be good to summarize how you plan to implement this before doing everything, that's way easier for both of us. I'm not trying to be super picky, but as I'll probably be the main maintainer of this long-term _and_ have plans for numpy.org (as you know), this setup is actually important.

mattip:
Closing after disucssion with @rgommers, the layout should as below. This still leaves the numpy organization with 4 html-documentation repos (3 once we fold neps into devdocs) dedicated to github pages.

| Content repo | Trigger | Destination Repo | Web link | Notes |
|---|---|---|---|---|
| numpy.org | Merge to `master' | numpy.github.com | https://numpy.org/ |
| numpy | Merge to `master` | devdocs | https://numpy.org/devdocs |
| numpy | Merge to `master` | neps | https://numpy.org/neps | Can be folded into devdocs |
| numpy | Manual release step | doc | https://numpy.org/doc | TBD - will become the canonical home of non-HEAD versions' documentation |

PR #31, opened 2019-08-08
add link back to main page (instead of scipy.org)

mattip:
Continuation of #25


mattip:
Once this is merged we should rebuild the site so this and #25 take effect

rgommers:
Thanks!

> Once this is merged we should rebuild the site so this and #25 take effect

wasn't this auto-building now?

mattip:
There is no CI at all in this repo

rgommers:
oh okay, then I'm confusing things - sorry for the noise

PR #38, opened 2019-09-03
Change NumFOCUS logos from PNG to SVG

rgommers:
Note that the small one isn't used currently, but we may need it for the redesign and it took me a while to get it, so let's put it in here.


mattip:
Thanks @rgommers

PR #46, opened 2019-10-02
GitHub actions for newsite branch

Shekharrajak:
- Github action : deploy-action created.
- Github workflow "deploy" job will trigger the deploy-action
- The pattern of the URL is : `pr-<pr_number>-numpy.org-newsite.surge.sh`

So for this PR it is : http://pr-46-numpy.org-newsite.surge.sh

Advantages:

- Now reviewing the UI changes easy because no need to clone the branch and check the UI locally. It can be seen in the deployed URL.
- PR author doesn't have to create GIF or attach screenshot for the UI changes. Dynamic changes can be seen easily in the URL.
- Every PR will be having the deployed URL so we can see the transformation of the website easily.


Shekharrajak:
Tried to get PR number using : `export PR_NUMBER="$(echo ${{REF}} | cut -d'/' -f3)"`

and `IFS='/' read -r -a array <<< "$REF"`

.github/workflows/main.yml
```diff
@@ -0,0 +1,55 @@
...
+    name: Test run
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os:
+          - macOS-latest
+          - ubuntu-latest
+          - windows-latest
+    steps:
+      - uses: actions/checkout@master
```
joelachance: do we want to specify the `newsite` branch for this? Obviously it will need to be master when we merge all the changes.

.github/workflows/main.yml
```diff
@@ -0,0 +1,55 @@
...
+    name: Deploy on surge
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version:
+          - 10
+        python-version:
+          - 3.7
+    steps:
+      - uses: actions/checkout@master
```
joelachance: Same question here.

.github/workflows/main.yml
Shekharrajak: Yes, that has to be changed when we merge newsite branch to master.

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM node:10-alpine
```
joelachance: Do we need node alpine? We might be able to use another image (something smaller, if it exists), although it probably doesn't really matter :)

joelachance:
Looks good so far!

Dockerfile
Shekharrajak: node docker image is used because later in deploy script it is installing the surge package.

Shekharrajak:
Currently, I am getting issue in assigning variable values and in setting env variable inside of the docker container after extracting the PR number from the $github.ref env variable. 

I will clean the commits.

Shekharrajak:
Finally more than 50 try, it worked and it is able to extract the PR number from the github.ref variable inside the docker container (while running the bin/sh script).

It is getting deployed in https://pr-46-numpy.org-newsite.surge.sh

.github/workflows/main.yml
```diff
@@ -0,0 +1,55 @@
...
+          - macOS-latest
+          - ubuntu-latest
+          - windows-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: run test
+        env:
+          REF: ${{ github.ref}}
+        run: |
+          echo Checking github env variables:
```
Shekharrajak: This has to be removed and proper test should run (similar to the travis CI/CD)

Shekharrajak: This has to be removed and proper test should run (similar to the travis CI/CD)

.github/workflows/main.yml
```diff
@@ -0,0 +1,55 @@
...
+      - uses: actions/checkout@master
+      - name: deploy in surge
+        uses: ./.github/actions/deploy-action
+        id: deploy
+        env:
+          SURGE_TOKEN: ${{ secrets.SURGE_TOKEN}}
+          SURGE_LOGIN: ${{ secrets.SURGE_LOGIN}}
+          REF: ${{ github.ref}}
+      - name: Get the output time
+        run: echo "The output was ${{ steps.deploy.outputs.deployed-domain }}"
```
Shekharrajak: This can be removed later. This is just to check the deployed-domain URL.

Shekharrajak:
@joelachance @rgommers please review. If it is looking good then we can merge this branch to newsite for now.

.github/workflows/main.yml
rgommers: We don't really have tests. Maybe it can be checked that the Hugo build finishes without errors or warnings?

.github/workflows/main.yml
rgommers: +1 for changing it to `actions/checkout@newsite` now

Dockerfile
rgommers: It maybe doesn't matter here, but Alpine is one of the most problematic distributions for Python packages, because it's musl rather than glibc based so `manylinux` wheels don't work if they contain compiled code. Just FYI

rgommers:
This looks pretty clean to me, happy to see it merged.

One question about the Surge setup: is this PR all that is needed, or do we need to have some account with access credentials that need storing somewhere?

Shekharrajak:
> This looks pretty clean to me, happy to see it merged.
> 
> One question about the Surge setup: is this PR all that is needed, or do we need to have some account with access credentials that need storing somewhere?

Thanks for the review. To deploy the static contents we need to have `SURGE_LOGIN` and `SURGE_TOKEN` as environment variable. In github action we can add environment variables secretly in `settings->secret` tab. I have added my surge token and login mail ID. It can be changed in the future and it will not affect anything.

.github/workflows/main.yml
Shekharrajak: As far as I understand, checkout action need the tag of the repository not the branch. So it is basically ` action/checkout@<tags>` .I tried with `newsite` and got this error in log: https://github.com/numpy/numpy.org/pull/46/checks?check_run_id=249019844

.github/workflows/main.yml
Shekharrajak: I have removed this job for now. We can add jobs later as per the requirements.

Dockerfile
Shekharrajak: Thanks for the info.

.github/workflows/main.yml
WarrenWeckesser: My understanding is that the the `@` symbol used in an `action` specifies the version *of the action*: https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstepsuses

Shekharrajak: Thanks @WarrenWeckesser for the link. I am wondering why it is working with this`@master` and the base branch for the PR is `newsite`. [Here](https://github.com/actions/checkout) I see that we can use specific branch using 

```
- uses: actions/checkout@master
  with:
    ref: some-branch

```

WarrenWeckesser: I'm still figuring this out, so all I can say is that `uses: actions/checkout@master`  means you are, in effect, using the development version of the `checkout` action, i.e. the `master`branch of https://github.com/actions/checkout

Shekharrajak: Yeah! It looks like you are correct. Since it is in this format `{owner}/{repo}/{path}@{ref}`. and using `with` option we are customizing the branch, sha, release version. 

So I think basically it is using this PR branch and latest commit to run the action specified.

WarrenWeckesser: It looks like the default branch (or SHA) that is checked out is explained in https://help.github.com/en/articles/events-that-trigger-workflows#webhook-events.  In particular, see the `GITHUB_SHA` column of the tables for [`pull_request`](https://help.github.com/en/articles/events-that-trigger-workflows#pull-request-event-pull_request) and [`push`](https://help.github.com/en/articles/events-that-trigger-workflows#push-event-push) events.

Also, README.md for the `checkout` action says 

>By default, the branch or tag ref that triggered the workflow will be checked out. If you wish to check out a different branch, specify that using `with.ref`

joelachance:
Looks good! Feel free to merge.

PR #47, opened 2019-10-08
Add names of "buckets" under Ecosystem on the front page

rgommers:
Wanted to open this for discussion, to see if these are the right headings/buckets to use or if people have better ideas.

Haven't found the time to write this content unfortunately.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -13,7 +13,7 @@ params:
     sizes: [400,600]
   hero:
     # Main hero title
-    title: Numpy
+    title: NumPy
     # Hero subtitle (optional)
     subtitle: The fundamental package for scientific computing and data science with Python
     # Button text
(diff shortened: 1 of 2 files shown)


rgommers:
Some good feedback from @mkg33 was to emphasize also individual scientific domains. One way to do that is copy what https://julialang.org/ does: a "Scientific Domains" tab that then just a range of domains with some explanation and links. It manages to list eight distinct science domains in one paragraph.

rgommers:
Okay went with the "Scientific Domains" idea for now. Will merge so we can iterate easier.

rgommers:
We have CI, and it's green:)

@Shekharrajak something didn't quite work, output is:
```
Run echo "The deployed-domain is "
  echo "The deployed-domain is "
  shell: /bin/bash -e {0}
The deployed-domain is 
```
and nothing got deployed. maybe you can look into that?

Shekharrajak:
I am looking at this issue. Thanks for letting me know.

Shekharrajak:
Fixed athttps://github.com/numpy/numpy.org/pull/58

PR #49, opened 2019-10-15
Easier build

stefanv:


rgommers:
I suspect that the `resources/` should not have been deleted, see gh-50.

Issue #50, opened 2019-10-17
Hugo build for newsite is broken 


rgommers:
Multiple people reported that the links on the main page seem to loop back to the front page when looking at the site with `hugo server`. I see that too, and rendering into `public/` by running `hugo` shows that the content of `index.html` is identical to that in `install/index.html` and other pages. 

@joelachance did this work for you, or are you seeing the same?

My Hugo is `0.58.3`, install with my linux distro package manager.

rgommers:
Ah it may have something to do with https://discourse.gohugo.io/t/hugo-extended-needed-for-hugo-fresh-theme/13844/25

Still figuring out what Hugo Extended means.

rgommers:
Going back to before gh-49 doesn't fix anything for me. Although it does seem to me that part of that PR may have been incorrect, due to (quote from discourse link above):

_We have a policy in which Hugo themes assets should be considered as stable and theme authors who use Hugo Extended features should commit their /resources/ directory so that a theme works in all versions of Hugo._

rgommers:
I actually have the latest extended version:
```
$ hugo version
Hugo Static Site Generator v0.58.3/extended linux/amd64 BuildDate: unknown
```

rgommers:
Tried to update to latest `master` of the Fresh theme, didn't help.

rgommers:
It's this issue: https://github.com/StefMa/hugo-fresh/issues/70#issuecomment-537851743

PR #51, opened 2019-10-17
Fix navigation to other pages with recent Hugo.  Closes gh-50

rgommers:
Also updates to latest version of the Fresh theme.

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -12,13 +12,20 @@
     </div>
 
     {{ block "main" . }}
+
+    {{ partial "single/single.html" . }}
+
+    {{ if .Params.include_footer }}
+    {{ partial "footer.html" . }}
     {{ end }}
 
     <!-- Back To Top Button -->
     <div id="backtotop"><a href="#"></a></div>
 
-    {{ if .Site.Params.sidebar }}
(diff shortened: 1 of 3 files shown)


rgommers:
Okay I identified and fixed the navigation issue, but the fix is incomplete and CI is right to explain. The second commit could be wrong, it's still failing:
```
Error: Error building site: failed to render pages: render of "page" failed: "/app/layouts/_default/single.html:13:34": execute of template failed: template: _default/single.html:6:7: executing 
"_default/single.html" at <partial "css.html" .>: error calling partial: "/app/layouts/partials
/css.html:13:34": execute of template failed: template: partials/css.html:13:34: executing 
"partials/css.html" at <toCSS $cssOpts>: error calling toCSS: type map[string]interface {} not 
supported in Resource 
```

I don't actually know what I'm doing with this theming, so I'll leave it here for now. @joelachance if you could take this over, that would be awesome.

joelachance:
I will look at this! Thank you for finding this fix, I hadn't had much luck with it.

stefanv:
I am also running v0.58.3, but the build works for me.

rgommers:
superseded by gh-59, closing

PR #52, opened 2019-10-17
Testing - Github Action: Deploy

Shekharrajak:
Intention is to check the URL :http://pr-52-numpy.org-newsite.surge.sh


Shekharrajak:
Oops, it looks like I rebase it incorrectly. I will remove other commits.

PR #58, opened 2019-10-19
Testing - Surge deploy

Shekharrajak:
Should get deployed at http://pr-58-numpy.org-newsite.surge.sh/


Shekharrajak:
There is 2 scripts : build.sh to generate the build and deploy.sh to deploy it in surge but after the build command inside docker container next commad is not recognizing the` ./public` folder.

Shekharrajak:
When I build the newsite branch then I found that it is building following files  

```
css         images      index.xml   js          sitemap.xml

```

That should be html file instead. @joelachance Please let me know if you already know about it.

joelachance:
@Shekharrajak, I don't know about the index.xml, but I'll take a look at it tomorrow.

joelachance:
@Shekharrajak, check this out:

https://stackoverflow.com/questions/42470517/hugo-not-generating-index-and-404-html-files

According that post, it seems we may need to build the submodules. I'm thinking we may want to try `git submodule update --init` (the command in the README) in line 2 of `build.sh`.

.scripts/build.sh
```diff
@@ -1,3 +1,6 @@
 #!/bin/sh
```
joelachance: Should we delete this file since its not being used, or do we still plan to move the build steps here?

joelachance:
Should be working now! Looks good to merge.

.scripts/build.sh
Shekharrajak: I will make a note. Actually previously in this PR: https://github.com/numpy/numpy.org/pull/46 , I used to call the `build.sh` and `deploy.sh` separately. But now we were facing [issue](https://github.com/numpy/numpy.org/pull/58#issuecomment-544115089) so I make it in one script file.

Shekharrajak:
Thanks @joelachance

Shekharrajak:
I am merging it if no one has any comments on this PR by tomorrow.

rgommers:
LGTM, merging. Thanks @Shekharrajak!

.scripts/build.sh
joelachance: Got it! I did see your comment in deploy.sh after I made this comment. Maybe we just delete or add a `#deprecated / not being used` comment in build.sh?

PR #59, opened 2019-10-20
Bug/50 broken links

### LINKED ISSUE #50: Hugo build for newsite is broken

joelachance:
A couple bonus commits here, but this accomplishes the following:

* Removes particles.js so we can feature some code in on the front page
* Fixes the broken links.

Turns out the fix in that issue was correct, but re-cloning `hugo-fresh` didn't work. I'm unsure when the cut releases over there, but even though that PR was merged on their side, it doesn't seem to be in their latest.

Resolves #50


config.yaml
```diff
@@ -1,6 +1,6 @@
 baseURL: http://numpy.org/
 languageCode: en-us
-title: Numpy.org
+title: Numpy
```
stefanv: NumPy?

joelachance: Should this be NumPy? I can change this.

layouts/index.html
```diff
@@ -0,0 +1,40 @@
+{{ define "main" }}
+    {{ if .Site.Params.hero }}
+    {{ partial "hero.html" . }}
+    {{ end }}
+
+    {{ if .Site.Params.section1 }}
+    {{ partial "section1.html" . }}
```
Shekharrajak: I think it is better to write better file name like instead of `section1.html` we can write `section_section_title_name.html`.

layouts/partials/hero-body.html
```diff
@@ -4,9 +4,33 @@
...
+                <code>
+                    def sample(some_list):
+                        print(some_list)
+                </code>
+            </div>
+        </div>
+    </div>
+    
+    <!-- <div class="container">
         <div class="columns is-vcentered">
```
Shekharrajak: comments can be removed completely.

layouts/partials/hero.html
```diff
@@ -0,0 +1,6 @@
+<section class="hero">
+  {{ partial "navbar.html" . }}
+  {{ partial "navbar-clone.html" . }}
+  {{ partial "hero-body.html" . }}
+  {{ partial "hero-footer.html" . }}
+</section>
```
Shekharrajak: Can we rename the ` hero.html` file name according to numpy

config.yaml
rgommers: yes, it's NumPy

rgommers:
Nice. All works for me, modulo the open comments this looks good to merge.

layouts/index.html
joelachance: `section1` is a hugo-fresh convention. It's only seen on the homepage, and it's the first section after the hero. We can rename it, but I think the naming convention you proposed is too much since this is only on the homepage. Ideas for a new name?

layouts/partials/hero.html
joelachance: This is again only showing up on the homepage. The above the fold section of the homepage is usually called the `hero` or `jumbotron`, so this is yet another convention. We can rename it, but this should only occur one time on any given site.

layouts/partials/hero-body.html
joelachance: removed

layouts/partials/hero.html
rgommers: Makes sense to not change theme naming conventions like `hero` or `section1`.

rgommers:
One thing I noted is that pages don't have the footer, and the header also doesn't contain the NumPy name or a way to navigate back to the main page.

joelachance:
@rgommers These are both things we can add! I'll make note and make those changes.

PR #60, opened 2019-10-22
Updating the navbar

InessaPawson:
Let's update the main navigation according to the approved sitemap (app.flowmapp.com/projects/84515/sitemap/).


rgommers:
Adds missing link to contribute page, looks good. Not sure this order is optimal (e.g. learn and documentation would make sense to have next to each other), but let's worry about that later. Merging, thanks @InessaPawson

Issue #48, opened 2019-10-10
Uniformity in Hugo and Sphinx template design


InessaPawson:
Uniformity in Hugo and Sphinx template design of numpy.org is desired. At the very least the same fonts and color-scheme (https://user-images.githubusercontent.com/98330/63642399-51f41480-c673-11e9-8d22-20315b71c83e.jpg) must be used.

joelachance:
@Shekharrajak , hoping we can pair on this one.

Here's what I've been using for the website, which I'll admit is probably different than what Inessa linked in the description and probably what you're using in the updated Sphinx theme.

Some of the colors and fonts I've been using (mostly coming from http://www.flatuicolorpicker.com/)
#013243 (dark blue)
#ffc553 (gold/yellow)
#ececec, #eee  (off-white)
#fff (white)
#6c7a89  (gray)
<img width="569" alt="Screen Shot 2020-03-20 at 8 10 55 AM" src="https://user-images.githubusercontent.com/3891660/77166501-5b672300-6a82-11ea-9ade-3b74fdd76996.png">


Fonts:
Lato https://fonts.googleapis.com/css?family=Lato:400,900&display=swap
Source Code Pro (for shell) https://fonts.googleapis.com/css?family=Source+Code+Pro&display=swap

I know we discussed Ubuntu in the past, and felt Lato looked a touch sharper when going back and forth. What's being used for the docs?

InessaPawson:
@joelachance #ffc553 (gold/yellow) is spot on, it's one of the NumPy official branding colors. I like the use of the various shades of gray too. 

Regarding the choice of font, an elegant and modern font, Ubuntu doesn't seem to enhance the UI of numpy.org. I like the use of Lato better.

joelachance:
@InessaPawson agreed, I feel the same. 

Regarding the colors, I do think some of the 'less bright' colors modernize things a bit while keeping the color scheme. I'm happy to do whatever here. Curious what was used for the Sphinx refresh @Shekharrajak!

Shekharrajak:
Hi,

While choosing color scheme,  I looked at previous site and took the one which suits the most (in my opinion, since it was not discussed that time). I used : #4dabcf , white, #404040 (in some background), #d9f4ff (for code highlight background color). 

Need to do changes accordingly. Tracking it here: https://github.com/scipy/scipy-sphinx-theme-v2/issues/30

joelachance:
[...]

PR #71, opened 2019-11-15
Feature/code demo

### LINKED ISSUE #66: Interactive Shell
Use Binder to create an interactive shell on the homepage: ThebeLab + MyBinder: https://thebelab.readthedocs.io/en/latest/

joelachance:
Resolves #66

### DIFF
--- a/layouts/partials/css.html
+++ b/layouts/partials/css.html
@@ -21,5 +21,6 @@
 <!-- Google Fonts - Ubuntu -->
 <link href="https://fonts.googleapis.com/css?family=Ubuntu&display=swap" rel="stylesheet">
 <link href="https://fonts.googleapis.com/css?family=Rubik&display=swap" rel="stylesheet">
+<link href="https://fonts.googleapis.com/css?family=Lato:400,900&display=swap" rel="stylesheet">
 <!-- Tabs CSS -->
 <link rel="stylesheet" type="text/css" href="{{ "css/tabs.css" | relURL}}" />
--- a/layouts/partials/css.html
+++ b/layouts/partials/css.html
@@ -22,5 +22,6 @@
 <link href="https://fonts.googleapis.com/css?family=Ubuntu&display=swap" rel="stylesheet">
 <link href="https://fonts.googleapis.com/css?family=Rubik&display=swap" rel="stylesheet">
 <link href="https://fonts.googleapis.com/css?family=Lato:400,900&display=swap" rel="stylesheet">
+<link href="https://fonts.googleapis.com/css?family=Source+Code+Pro&display=swap" rel="stylesheet">
 <!-- Tabs CSS -->
 <link rel="stylesheet" type="text/css" href="{{ "css/tabs.css" | relURL}}" />
--- a/config.yaml
+++ b/config.yaml
@@ -17,7 +17,7 @@ params:
     # Main hero title
     title: NumPy
     # Hero subtitle (optional)
-    subtitle: The fundamental package for scientific computing and data science with Python
+    subtitle: The fundamental package for scientific computing with Python
     # Button text
     buttontext: Get started
     # Where the main hero button links to
--- a/layouts/partials/hero-body.html
+++ b/layouts/partials/hero-body.html
@@ -27,7 +27,6 @@
                 <div class="demo-title">Try NumPy!</div>
                 <div class="demo-copy">Give NumPy a shot. We should write some better copy here.</div>
             </div>
-            <div class="demo-caret">></div>
                 <pre data-executable="true" data-language="python">
 %matplotlib inline
 import numpy as np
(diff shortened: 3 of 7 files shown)


stefanv:
@joelachance This is great!  Should we consider a more minimalistic environment with only NumPy installed, and perhaps initialize the terminal with numpy already imported as np?

rgommers:
I like this as well! It's quite responsive, much faster than I'm used to with Binder. I guess it's a function of the set of requirements.

> Should we consider a more minimalistic environment with only NumPy installed, and perhaps initialize the terminal with numpy already imported as np?

Is this for loading speed, or for saving Binder resources, or ... ?

stefanv:
> Is this for loading speed, or for saving Binder resources, or ... ?

It probably won't affect the loading speed too much; more so we have control over the environment we operate in (numpy version, etc.).

If we want to increase response time, we do have the option of hosting (or sponsoring) a MyBinder instance that serves the NumPy image exclusively (most of the start-up time on the MyBinder time is spent moving images to execution nodes, if I understand correctly).

There are further optimizations on the table, such as holding a cache of ready kernels and dishing them out as needed, but that is clearly out of scope for this project.

rgommers:
> It probably won't affect the loading speed too much; more so we have control over the environment we operate in (numpy version, etc.).

Adding just `matplotlib` seems nice, that way we can use nicer demo code. That should not affect the numpy version that gets installed.

And perhaps we can put the `%matplotlib inline` into a config file to always enable it?

rgommers:
@scopatz this is the PR with ThebeLab that we talked about. @joelachance, @scopatz is working on something similar, more general / not Jupyter based (he can explain better). He may be interested to chat/collaborate.

scopatz:
Thanks @rgommers!

@joelachance - can we maybe jump on a call next week sometime? Feel free to email me at scopatz(AT)quansight.com

joelachance:
> This is great! Should we consider a more minimalistic environment with only NumPy installed, and perhaps initialize the terminal with numpy already imported as np?

What we currently have in this PR is the default ThebeLab/Binder image. We can create our own if we'd like, but we will have to host it somewhere. The default image is already very minimal, but as it does take a brief moment to initialize, a smaller image might improve loading time (probably not a huge priority).

A couple of enhancements I think would be nice are as follows:
* A spinner/loader inside of the 'Run' button (in place of the 'Run' text) when the kernel is initializing
* A default numpy import at the top, maybe even a static one above the cursor. I considered removing imports from the default example (but including them when the code is run), but I'm not sure this is best. I think it's good for beginners to know you have to import the library. @rgommers, I agree about including `%matplotlib inline` with the code that is run, but not in the terminal window.

There's one or two more, but I can't remember right now :) I'm on the fence about including `matplotlib`, mostly because of css/styling reasons (the default plot works, but the 'terminal window' isn't going to resize for a larger plot, for example). I think `seaborn` might be included too, but I haven't checked.

joelachance:
@scopatz, I will email you!

layouts/partials/css.html
```diff
@@ -21,5 +21,7 @@
 <!-- Google Fonts - Ubuntu -->
 <link href="https://fonts.googleapis.com/css?family=Ubuntu&display=swap" rel="stylesheet">
 <link href="https://fonts.googleapis.com/css?family=Rubik&display=swap" rel="stylesheet">
```
Shekharrajak: If you are using Lato font now then can we remove Rubik font link ?

joelachance: Yes, we can.

I've been playing around with fonts for the main hero title. I'm not sure Ubuntu is the best choice here, although I do think it looks quite nice for body text.

I'll write more soon in #48.

rgommers:
This is in good shape so I'm merging it to not have to migrate it to `master` later. The left over review comments can be discussed/addressed here of course (and fixed with a PR to `master` :) )

rgommers:
Thanks @joelachance and all reviewers, this is really nice!

InessaPawson:
@joelachance I'm submitting the screenshots of the shell to follow up on our last meeting discussion. I'm using Mac Pro. It's the same scenario with Safari and Firefox. I'm unable to get past the third screenshot. It was not an issue while I was in the US.

numpyshell1
numpyshell2
numpyshell3

joelachance:
@InessaPawson The third screenshot is the shell itself-- I had restyled it, but you're the second person to note some confusion about the new styles, so I'm planning on updating this to look more like it did before. If you click at the end of the `import numpy as np` line, you should be able to type more and run your code.

InessaPawson:
A-ha! I guess, my confusion is exacerbated by the fact that I remember the previous version.:) Perhaps, we could  offer a prompt in a darker shade of grey, e.g. “Enter your code here…”
Or something similar to [repl.it](https://repl.it/repls/StimulatingMediumpurpleDisassembly) : “Not sure what to do? Run some examples/an example. Start typing to dismiss.” 
P.S. You have to appreciate the randomly generated name of this repo.;)

PR #79, opened 2019-11-24
Another try at fixing the Surge deploy, using the correct baseURL

rgommers:
Should be deployed at http://numpy-79-numpy.org-newsite.surge.sh


rgommers:
Nope, still getting:
```
PR_NUMBER: 79
SURGE_LOGIN: 
SURGE_TOKEN: 
```
rather than getting `***` for login and token.

rgommers:
still didn't work with last commit

.scripts/deploy.sh
```diff
@@ -4,21 +4,30 @@
 # NOTE: Running build.sh and then deploy.sh is not finding the public
 # folder which is generated after the build command (hugo).
 git submodule update --init
+
+# Find PR number and determine where to deploy to with GitHub Actions
+PR_NUMBER="$(echo $REF | cut -d'/' -f3)"
+DOMAIN_NOHTTP="numpy-${PR_NUMBER}-numpy.org-newsite.surge.sh"
```
joelachance: Should we update this name to remove `newsite`? We could probably remove the `numpy.org` before it as well, if you'd like.

rgommers: Yes, I just wanted to check whether it worked before changing the name. It doesn't at the moment, and I'm not sure if the name has anything to do with it. @Shekharrajak any thoughts?

Shekharrajak: URL can be anything. It is not a problem.

joelachance: I think the issue is with Hugo and how its relative & absolute URL's work. I don't think there's a deploy script/surge issue (that's my intuition, anyway).
@Shekharrajak/@rgommers, I had updated urls to relative URLs, but I don't think that worked. Any thoughts/suggestions here?

rgommers: @joelachance I'm fairly sure that's unrelated to this. If it doesn't deploy to Surge in the first place because the secret tokens don't show up, we never get to the point where absUrl/relUrl matters

joelachance: Ah, I thought it had been loading. You're right.
I think it may have been working previously, but it takes a *very* long time to load sometimes...

stefanv:
Just double checking, but are you checking the token values after merging?  Because you won't see it on a PR:

"Secrets are not passed to workflows that are triggered by a pull request from a fork of this [numpy/numpy.org] repository."

rgommers:
> "Secrets are not passed to workflows that are triggered by a pull request from a fork of this [numpy/numpy.org] repository."

Ah, that may explain it. That makes the feature fairly useless though ...

rgommers:
So the fix is to use the `secrets` context: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets#using-encrypted-secrets-in-a-workflow

Shekharrajak:
> So the fix is to use the `secrets` context: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets#using-encrypted-secrets-in-a-workflow

Yes, thanks for sharing the link.  I have read about it. But this example itself add `secrets`, I think I should try it out directly passing the encrypted string to `env` option in workflow. I will try it out.

Shekharrajak:
I am communicating to surge and github team about this. Github support replied with : 

"
This is because a user could potentially modify your workflow to output or expose the secret by simply creating a pull request.

These secrets will be available again once this pull request is merged in to your upstream repository.
"

But this should not happen because it work for my PRs (and worked on #71 as well). 

I am looking into it.

Shekharrajak:
Should we go for travis , because I don't think that out `secret` variables are greater than 64 KB :  https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets#using-encrypted-secrets-in-a-workflow

rgommers:
GitHub Actions indeed seems unable to solve this issue: https://github.community/t5/GitHub-Actions/Allow-secrets-to-be-shared-with-trusted-Actions/td-p/34278. 

Happy to go with TravisCI

rgommers:
I'll close this PR, since it's clearly not going to work.

Shekharrajak:
Thanks @rgommers for sharing the link. Now we can track what is going on.

PR #97, opened 2019-12-07
Feature/67 footer

### LINKED ISSUE #67: Update Footer Layout
Update the footer layout. See the below image (keep 'old' NumPy logo): 
Screen Shot 2019-11-10 at 8 51 52 PM

Question: how should an email list function? Should these be written to S3? Emailed to someone?

joelachance:
Resolves #67

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -62,19 +62,6 @@ params:
       url: /
       buttonText: Get started
   section5: false
-  footer:
-    # Logo (from /images/logos/___)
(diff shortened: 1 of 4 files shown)


rgommers:
This looks awesome:

image

The hover effect on the social media icons is a nice touch:) And the "thank you when you add an email address as well:)

All works as expected too, so in it goes. Thanks @joelachance! And thanks @InessaPawson for the design!

PR #99, opened 2019-12-11
Update Twitter handle to `numpy_team`

rgommers:
And we're live: https://twitter.com/numpy_team

No tweets yet, will fix that before we launch the website, as well as set up access for other people and some kind of guideline on what we tweet (Dask just figured that out, we can copy from them).


rgommers:
Another trivial one-line commit, so merged straight away

Issue #103, opened 2019-12-13
Content: ML Tab


shaloo:
Add content for Machine Learning tab under Ecosystem on landing page

shaloo:
Updated- refer to PR https://github.com/numpy/numpy.org/pull/106

shaloo:
<img width="925" alt="Screenshot 2019-12-13 at 8 02 39 PM" src="https://user-images.githubusercontent.com/5890484/70807655-fe553480-1de3-11ea-967a-abb8d3ed7327.png">

rgommers:
this is done, closing

PR #106, opened 2019-12-13
Fixes #gh-103 content for tab machine learning added

shaloo:
#### Deployed URL

https://pr-<pr_number>-numpy.surge.sh


shaloo:
This is how the draft shows up on localhost:1313 (the LHS gif is animated - this png does not show that btw)

Screenshot 2019-12-13 at 8 02 39 PM

rgommers:
The logos at the bottom are huge when I try this PR locally:

image

I think it's probably better to leave out those logos. Note that Chainer and Keras are no longer actively developed, and XGBoost shows up twice.

layouts/partials/tabs.html
```diff
@@ -295,22 +295,72 @@ <h3 class="subtitle is-5 is-muted">{{ $subtitle }}</h3>
...
-	Other suggestions for Machine Learning:
-
-        Visualization: Yellowbrick, Eli5
-        ML libraries: XGBoost, LightGBM, CatBoost
-        AI libraries: PyTorch, MXNet
-        
-	</p>
+       	   <div class="grid-container">
+	      <div>
+		      <a href="https://ai.googleblog.com/2016/12/open-sourcing-embedding-projector-tool.html"> <img src="images/content_images/ml_img/pytorch-ml-anim.gif" alt="pytorch figure" align="middle" style="border-radius: 10px"></a>
```
rgommers: The file name and alt text are confusing - this was made with Tensorflow not PyTorch. Could you change that?

Also, is this okay to include copyright-wise?

layouts/partials/tabs.html
```diff
@@ -295,22 +295,72 @@ <h3 class="subtitle is-5 is-muted">{{ $subtitle }}</h3>
...
-	</p>
+       	   <div class="grid-container">
+	      <div>
+		      <a href="https://ai.googleblog.com/2016/12/open-sourcing-embedding-projector-tool.html"> <img src="images/content_images/ml_img/pytorch-ml-anim.gif" alt="pytorch figure" align="middle" style="border-radius: 10px"></a> 
+              </div>
+	      <div>
+                 <p>
+		 <b>Machine learning</b> (ML) enables computers to learn using
+		    data, without having to be explicitly programmed.
+		    <b>NumPy</b> is the foundation of all data pre-processing
```
rgommers: This is not quite true. Perhaps better to say that these libraries all have data structures that are inspired by NumPy.

layouts/partials/tabs.html
```diff
@@ -295,22 +295,72 @@ <h3 class="subtitle is-5 is-muted">{{ $subtitle }}</h3>
...
+	   </div>
+	   <div>
+	      <div>
+               	 <p>
+		    <a href="https://www.tensorflow.org">Tensorflow’s</a>
+		    deep learning capabilities help to define and run
+		    computations involving tensors. Tensors have broad
+		    applications in Speech and image recognition, Text-based
+		    applications, Time-series analysis and video detection.
+		    <a href="https://keras.io">Keras</a> provides high level
```
rgommers: Keras is no longer maintained as a standalone library, so let's not mention it here.

layouts/partials/tabs.html
```diff
@@ -295,22 +295,72 @@ <h3 class="subtitle is-5 is-muted">{{ $subtitle }}</h3>
...
+		    neural network APIs. <a href="https://pytorch.org">PyTorch
```
rgommers: I'd suggest a more equal distribution of text between Tensorflow and PyTorch. The latter is already more popular with researchers than the former. Also JAX deserves a brief mention.

rgommers: And let's mention MXNet here rather than in the next paragraph, that fits better (in scope it's similar to Tensorflow and PyTorch).

layouts/partials/tabs.html
shaloo: The image is linked to original one.  I am mentioning source as well so there should be no copyright issue (I think).  I have copied the image for fast loading (as done in other tabs)

layouts/partials/tabs.html
shaloo: ok removed

layouts/partials/tabs.html
shaloo: ok done.

shaloo: Removed logos, edited text as suggested.
Screenshot 2019-12-19 at 11 29 27 AM

rgommers:
The tab is looking pretty good now. This PR still has all the images in it though, those should be removed. More importantly, the animation that's still used in the tab is a 9 Mb file. That seems like too much, so we'll need to decide how to deal with that - perhaps the size can be reduced by a lot (at least 10x I'd say), or otherwise we should switch to a static image.

shaloo:
Oh! Why are the logos showing?  I don't see any in my branch - double checked - no file commit got missed out.  Let me check again.

shaloo:
Pls review and merge. Thanks.

shaloo:
@rgommers, is this good to go now with reduced animation size?

rgommers:
Yes it is, thanks @shaloo!

There was one open comment left, but I'll take that along when I do an edit of all content.

PR #113, opened 2019-12-21
Create report-handling-manual.md

InessaPawson:

### DIFF
--- a/content/en/code-of-conduct.md
+++ b/content/en/code-of-conduct.md
@@ -59,7 +59,7 @@ If your report involves any members of the Committee, or if they feel they have 
 
 ### Incident reporting resolution & Code of Conduct enforcement
 
-_This section summarizes the most important points, more details can be found in_ [NumPy Code of Conduct - How to follow up on a report](https://numpy.org/devdocs/dev/conduct/report_handling_manual.html).
+_This section summarizes the most important points, more details can be found in_ [NumPy Code of Conduct - How to follow up on a report](/report-handling-manual).
 
 We will investigate and respond to all complaints. The NumPy Code of Conduct Committee and the NumPy Steering Committee (if involved) will protect the identity of the reporter, and treat the content of complaints as confidential (unless the reporter agrees otherwise).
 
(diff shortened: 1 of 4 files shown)


rgommers:
I had to think about it for a little, but I agree that putting this page on numpy.org is the right thing to do. After the site goes live, we can remove it from the developer guide, and link here instead.

rgommers:
@joelachance in this PR I added CSS for list rendering (`<ul>`, `<li>`), which makes all the pages look a lot better. Unfortunately it messes up the footer styling, which relied on unstyled lists. To fix that I can do something like:
```
diff --git a/assets/fresh/partials/_footer.scss b/assets/fresh/partials/_footer.scss
index 74ced3f..be9fa37 100644
--- a/assets/fresh/partials/_footer.scss
+++ b/assets/fresh/partials/_footer.scss
@@ -24,6 +24,8 @@ footer.footer-dark {
         ul.link-list {
             line-height: 40px;
             font-size: 1.1rem;
+            padding: 0px;
+            margin-left: 0px;
             a {
                 color: $light-blue-grey;
                 font-weight: 400;
@@ -33,6 +35,11 @@ footer.footer-dark {
                 color: $smoke-white;
             }
         }
+        li.link-list {
+            list-style-type: none;
+            padding: 0px;
+            margin-left: 0px;
+        }
         .level-item .icon {
             color: $secondary;
             transition: all 0.5s;
@@ -41,4 +48,4 @@ footer.footer-dark {
             }
         }
     }
-}
\ No newline at end of file
+}
```
however that is in the theme. What do you think the best way to apply a patch like that is? Modifying a submodule isn't right I think, so should we move that whole `_footer.scss` to `assets/fresh/partials/` to have it overwrite the theme default?

joelachance:
Ok, I've pushed a small change here that gives it a more specific rule and doesn't break the footer. I've avoided attempting to change any of the sass files in the theme and just give all styling updates more specific rules.

All the content is rendered via the `content.html` page via `layouts/partials/single/content.html`.

Hope this helps!

rgommers:
Excellent, thanks Joe! In it goes.

PR #115, opened 2019-12-22
Create team.md

InessaPawson:
Adding content, @joelachance is working on the rest.

### DIFF
--- a/content/en/team.md
+++ b/content/en/team.md
@@ -1,0 +1,15 @@
+---
+title: Team
+ 
+sidebar: false
+
+---
+
+*TO DO: Add a Google Maps snapshot illustrating where members of the NumPy team are from.*
+
+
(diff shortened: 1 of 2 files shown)


content/en/team.md
```diff
@@ -0,0 +1,15 @@
...
+title: Team
+ 
+sidebar: false
+
+---
+
+*TO DO: Add a Google Maps snapshot illustrating where members of the NumPy team are from.*
+
+
+We are an international team on a mission to support scientific and research communities worldwide by building quality, open-source software. [**Join us**](https://numpy.org/contribute.md)!
```
rgommers: This link won't work, it should be `[**Join us**](/contribute)`.

With Nikola, starting with `/` will create a link relative to the root of the domain, while leaving off the `/` means relative to the current page.

rgommers: pushed a commit with a fix

rgommers:
Merged, thanks @InessaPawson

PR #117, opened 2019-12-23
Content shortcuts

### LINKED ISSUE #88: Content Page Layout
- [x] Create a sidebar with shortcuts to headers inside the page
- [x] Add a 'Get Help?' link below the shortcuts
- [x] Add the footer to the content pages

The 'Get Help?' link should only render on the following pages:

- Install
- Documentation
- Learn
- Community
- Contribute

joelachance:
Resolves #88


static/css/styles.css
```diff
@@ -381,4 +381,53 @@ p {
...
+
+/* Shortcuts */
+.content-padding {
+    padding: 3rem 1rem;
+}
+
+.shortcuts-container {
+    width: 100%;
+    min-width: 10vw;
+    max-width: 15vw;
```
rgommers: This is the reason for the large whitespace in the sidebar. min `5vw`, max `10vw` looks better I think

joelachance: We can update that!

The shortcuts will scroll to those header sections, and aren't links.

rgommers: Oh yes they do scroll, interesting effect!

Okay, if you push a tweak for the whitespace I'll merge it straight away

joelachance: If you can think of anything to make it obvious those scroll to their respective sections, let me know.

rgommers: I think it's fine as is, I just wasn't seeing it on that page because the whole page fits in a single browser window hence nothing happened. Not styling it as a link (different font color) seems like a good idea.

rgommers:
In it goes, thanks Joe!

rgommers:
Oh, now that I looked at gh-88 again: we wanted to control which pages have a sidebar (e.g. `arraycomputing` didn't need "Get Help"). Not sure how easy that is. The frontmatter for each page content file has
```
sidebar: false
```
in it, but I think that refers to something else (can't remember)?

joelachance:
That does refer to something else. It shouldn't be too hard, but unfortunately it's probably a little more config. I can create a ticket and put a fix in!

rgommers:
Sounds good. It's something we could leave till after the first release if it takes too much time.

PR #125, opened 2020-01-04
Remove "general purpose" tab

rgommers:
I had too much trouble figuring out what should go in that tab,
which is a sign that it probably shouldn't exist.

### DIFF
--- a/layouts/partials/tabs.html
+++ b/layouts/partials/tabs.html
@@ -49,12 +49,6 @@
             </a>
             </li>
 
-            <li>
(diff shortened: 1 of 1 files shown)


rgommers:
Let's deal with the tab styling (width of tabs not matching the width of the text box under it) later. This was already wrong, and is still wrong after this PR.

rgommers:
@shaloo apologies, I should have done this sooner.

rgommers:
Two thumbs up, merging

PR #126, opened 2020-01-04
Update the About Us page

rgommers:
added logos, teams, steering council members, and the donate link


rgommers:
Note that the logos aren't in a nice gallery, they're just one after the other. I'm not sure how to best do this, so I thought I'd first get all the content in.

Same for the donate link: would be good to style that as a button similar to the Download/Run buttons on the front page.

rgommers:
In this goes. Making the logos look nice is still a TODO

PR #14, opened 2018-09-13
MAINT: add Numeric manual link

tylerjereddy:
@stefanv asked me to add a historic link to the `Numeric` manual re: being helpful for citing in the SciPy manuscript.

PR #133, opened 2020-01-07
Update link to Numeric manual, remove pdf from repo

rgommers:
Reason: the download link isn't stable anyway due to a Sphinx bug (see gh-14), so no reason to keep this pdf in the repo.

(note: discovered in the publication process of the SciPy paper)

This is also nicer, since there's now correct metadata, a DOI, etc.


rgommers:
Since this has gone in the SciPy paper and it seems straightforward, I'll merge it so I can forget about this again.

PR #144, opened 2020-01-15
New Crowdin translations

Shekharrajak:
Enjoy!

PR #146, opened 2020-01-15
exit if there is error while running the docker image in travis CI

Shekharrajak:
Related to this: https://github.com/numpy/numpy.org/pull/144#issuecomment-574793172

### DIFF
--- a/.scripts/build.sh
+++ b/.scripts/build.sh
@@ -8,3 +8,4 @@ set -euo pipefail
 git submodule update --init
 hugo
 
+echo $?
--- a/.scripts/build.sh
+++ b/.scripts/build.sh
@@ -8,13 +8,6 @@ set -euo pipefail
 git submodule update --init
 hugo
 
-echo $?
-
-if [ $? -eq 0 ]
-then
-  echo "Success: build."
-  exit 0
-else
-  echo "Failure: Error while building" >&2
-  exit 1
-fi
+# If it is not present that means build failed.
+# docker failure returns non error exit code so travis doesn't come out.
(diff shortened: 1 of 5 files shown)


.scripts/install.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -euo pipefail
```
mattip: You might want to add `-x` which prints commands and their arguments as they are executed.

.scripts/build.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 
+set -euox pipefail
```
rgommers: Does this work? `pipefail` is connected to the `-o` option only. My bash isn't great, but I'm sure `set -euxo pipefail` works, not sure about this version

Shekharrajak: I think I need to create any build failure to verify it. in travis

Shekharrajak:
It looks like it didn't work

Shekharrajak:
related issue found: https://github.com/travis-ci/travis-ci/issues/8920

Shekharrajak:
Still docker is returning non error code to travis CI. I have tried various way but still no use. It looks like I have to drop the idea of docker container and use the travis yaml file for the configuration.

rgommers:
Yeah, Docker seems overkill - it's a few shell scripts that should be relatively straightforward, just `.travis.yml` should do it

Shekharrajak:
Removed the docker part : https://github.com/numpy/numpy.org/pull/166

Issue #101, opened 2019-12-11
Style: Loading icon disruptive


stefanv:
What do folks feel about removing the concentric circle blip that appears on loading each page?  I, for one, find it quite disruptive on every click / page load.

rgommers:
+1 from me

shaloo:
+1 for removing

joelachance:
+1 here as well. This is default hugo-fresh behavior, so we can update it.

joelachance:
What would be preferred to the current loader?

rgommers:
No animation at all, just load the page as fast as possible I'd say.

MarsBarLee:
Here's the website without the loading circle, as the original issue requested.
![WithoutAnyAnimations](https://user-images.githubusercontent.com/46167686/72574534-67bedc00-3897-11ea-8d90-6be1baca723f.gif)

Here's the website without any animations the loading circle or any fade-in animation, as Ralf suggested.
![WithoutLoadingFrontPage](https://user-images.githubusercontent.com/46167686/72574546-6db4bd00-3897-11ea-9e31-99399b30bd46.gif)

Here's the website without the loading circle but a fade-in animation only for the front page
![WithLoadingFrontPage](https://user-images.githubusercontent.com/46167686/72574554-73120780-3897-11ea-9f24-d71cc5641e50.gif)

I like third one the most, even if it's a little bit slower for that page. The first page feels too slow and the second one feels a little too adrupt.

What do you guys think?

joelachance:
Nice, well done! I also like the third one a lot, but I think I'd end up picking the second. The time between initial page request and usability (I think there's probably an acronym for this I'm forgetting) is most important on the homepage. I'll admit the fade doesn't *really* affect it, I think it'll appear to users as an extra unnecessary moment.

I'm splitting hairs here, but 2 or 3 I think would be great.

InessaPawson:
@MarsBarLee , thank you for going the extra mile by giving us several options. I believe most of the numpy.org visitors will more appreciate the speed that comes with the second option.

rgommers:
Nice @MarsBarLee, both options look great to me! Now that you have both already, you could also put both in with one enabled one disabled and a simple if-statement to switch between them (assuming that's easy now). Then we can try out the prettier option 3, and if it does indeed feel too slow we can just flip the switch.

stefanv:
[...]

PR #150, opened 2020-01-20
Removed loading animations

MarsBarLee:
Removed loading icon and fade-in animation before each page loads.

For the front-page, there is a if-statement for the fade-in animation. It's currently set to false.

[Issue 101](https://github.com/numpy/numpy.org/issues/101)

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -6,28 +6,26 @@
     {{ partial "css.html" . }}
 
     <!-- Refer: https://regisphilibert.com/blog/2018/08/hugo-multilingual-part-1-managing-content-translation/ -->
-    {{ if .IsTranslated }}
-      {{ range .Translations }}
-      <link rel="alternate" hreflang="{{ .Language.Lang }}" href="{{ .Permalink }}" title="{{ .Language.LanguageName }}">
-      {{ end }}
-    {{ end }}
+    {{ if .IsTranslated }} {{ range .Translations }}
+    <link
+      rel="alternate"
+      hreflang="{{ .Language.Lang }}"
+      href="{{ .Permalink }}"
+      title="{{ .Language.LanguageName }}"
+    />
+    {{ end }} {{ end }}
   </head>
-  <body>
-    <!-- Preloader -->
-    <div id="preloader">
-      <div id="status"></div>
-    </div>
 
-    {{ block "main" . }}
-    {{ end }}
+  <body>
+    <!-- Preloader. Set to true for front-page fade-in animation -->
+    {{ if false }}
+    <div id="preloader"></div>
+    {{ end }} {{ block "main" . }} {{ end }}
 
     <!-- Back To Top Button -->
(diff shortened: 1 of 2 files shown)


layouts/_default/baseof.html
```diff
@@ -6,28 +6,26 @@
...
     <!-- Back To Top Button -->
     <div id="backtotop"><a href="#" id="backtotop-color"></a></div>
 
-    {{ if .Site.Params.sidebar }}
-    {{ partial "sidebar.html" . }}
-    {{ end }}
-
-    {{ partial "javascript.html" . }}
+    {{ if .Site.Params.sidebar }} {{ partial "sidebar.html" . }} {{ end }} {{
+    partial "javascript.html" . }}
```
joelachance: IMO, I have a harder time reading this as a single line. I think nesting and using multiple lines is just fine here!
Same on LOC 9, 13, & 23.

rgommers:
@MarsBarLee this doesn't actually build for me, it fails with:
```
$ hugo
Total in 10 ms
Error: add site dependencies: load resources: loading templates: 
"/home/rgommers/code/websites/numpy.org/layouts/_default/baseof.html:28:1": parse master failed: 
template: index.html:28: unexpected unclosed action in command
```

If it does work for you, then it's probably due to using different Hugo versions. Mine is:
```
$ hugo version
Hugo Static Site Generator v0.62.0/extended linux/amd64 BuildDate: unknown
```

layouts/_default/baseof.html
rgommers: The line break at the end of line 28 probably causes my build failure. Undoing this change should fix it.

MarsBarLee: I realized that this problem is caused by my code editor and formatting. It moves the lines and white text around when I save. I'll change my settings and update my pull request without these problems.

joelachance:
This one loads for me. However, I did notice there's some 'flash' when the page first loads since we're still fetching fonts, etc. We can avoid this by doing this:

`{{ if .IsHome }}` instead of `{{ if false }}`.
source: https://discourse.gohugo.io/t/how-to-check-for-home-page/149/24

When trying that, I don't see the loading icon, but it does seem to fade in nicely. No loader on the other pages.

Otherwise looks great!

MarsBarLee:
> This one loads for me. However, I did notice there's some 'flash' when the page first loads since we're still fetching fonts, etc. We can avoid this by doing this:
> 
> `{{ if .IsHome }}` instead of `{{ if false }}`.
> source: https://discourse.gohugo.io/t/how-to-check-for-home-page/149/24
> 
> When trying that, I don't see the loading icon, but it does seem to fade in nicely. No loader on the other pages.
> 
> Otherwise looks great!

Ok, made those changes and pushed! If we decide to remove the loading entirely, then we'll remove the preloader section altogether.

rgommers:
First PR merged! Nice @MarsBarLee :)

PR #151, opened 2020-01-21
section2 2x2 tiles

### LINKED ISSUE #84: Homepage - "In a nutshell"
Change UI to the two column table.

joelachance:
Adds 2x2 tiles, a rollover effect, and updated the title to 'Key Features'. Removes icons (for now).

Resolves #84


joelachance:
I'd like to update the 'Ecosystem' header to look more like this one. Also, I hope the wording update from 'In a Nutshell' to 'Key Features' is welcomed, I liked how PyTorch had their 'key features' section.

If that sounds good, I say we merge and I'll fix the Ecosystem header in the tab PR.

rgommers:
Agreed, I like "key features". It does point at something a little different (e.g. open source isn't a key feature I'd say) so we may want to edit/extend a little.

PR #154, opened 2020-01-22
Feature/54 about us

### LINKED ISSUE #54: Style About Us Page
The new link doesn't have `salsalabs` in it anymore, and looks a lot more trustworthy: https://numfocus.org/donate-to-numpy

Needs to be updated on the front page.

joelachance:
Fixes the images on about us, adds a donate button at the bottom.

Resolves #54.

### DIFF
--- a/content/en/about.md
+++ b/content/en/about.md
@@ -32,7 +32,6 @@ Emeritus members are:
 - Alex Griffing (served: 2015-2017)
 - Marten van Kerkwijk (served: 2017-2019)
 
-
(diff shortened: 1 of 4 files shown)


joelachance:
Note: we'll want to add shortcodes to the other language pages once they're merged.

joelachance:
@rgommers That was not on purpose! That was an oversight, I'll add those back. Good catch.

joelachance:
Fixed!

PR #155, opened 2020-01-22
Feature/70 lang dropdown

### LINKED ISSUE #70: Select language dropdown
* We can have language options written in that language only. i.e. "Hindi " should be written in hindi language, "Dutch" should be written in dutch language and so on.

* We can use the respective country flags before the language option as well

joelachance:
The languages dropdown now updates every 2 seconds to say 'languages' in a translated/supported language.
Language names are also updated in the dropdown.

Someone should check my translations :) 

Resolves #70.

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -94,20 +94,3 @@ if (shortcutsTarget.length > 0) {
     })
   });
 }
-
-// Scrolling Languages Dropdown
-Array.prototype.recurse = function(callback, delay) {
-  var self = this
-  var index = 0;
-
-  setInterval(function () {
-    callback(self[index], index);
-    var next = index+1;
-    index = next < self.length ? next : 0;
-  }, delay);
-}
-
-const langs = ['Languages', 'Talen', 'बोली']
-langs.recurse(function(item) {
-  $('.navbar-link').html(`<span class="language-text">${item}</span>`);
-}, 2000);
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -94,3 +94,20 @@ if (shortcutsTarget.length > 0) {
     })
   });
 }
+
+// Scrolling Languages Dropdown
+Array.prototype.recurse = function(callback, delay) {
+  var self = this
+  var index = 0;
+
+  setInterval(function () {
+    callback(self[index], index);
+    var next = index+1;
+    index = next < self.length ? next : 0;
+  }, delay);
+}
+
+const langs = ['Languages', 'Talen', 'बोली']
+langs.recurse(function(item) {
+  $('.navbar-link').html(`<span class="language-text">${item}</span>`);
+}, 2000);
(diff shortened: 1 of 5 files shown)


static/js/app.js
```diff
@@ -93,4 +93,21 @@ if (shortcutsTarget.length > 0) {
...
+  var index = 0;
+
+  setInterval(function () {
+    callback(self[index], index);
+    var next = index+1;
+    index = next < self.length ? next : 0;
+  }, delay);
+}
+
+const langs = ['Languages', 'Talen', 'बोली']
```
rgommers: This will be hard to find, at least deserves a comment in `config.yaml`

rgommers:
What do others think?

joelachance:
I'm happy either way-- I thought this might be a nice feature. If people would prefer flags, I think that's great too.

I do think something is better than nothing here.

joelachance:
Interestingly, after a quick google for language flags, it seems like the consensus is to _not_ use flags. Most examples I'm seeing display the selected language, so we would render 'English' instead of 'Languages' at the top level. Inside the dropdown would be supported languages written natively. This seems clean and simple to me, vs. the javascript solution or adding flags. Anyone have thoughts?

MarsBarLee:
I initially agreed on the flags, but I also found advice online not to use them. 

Maybe the language section can change depending on the context, if the user is at the top of a page?

If a user is at the top of a page, the language section could change every 2 seconds to display the supported languages written natively, like Joe's original suggestion. The user can hover to display the dropdown menu that displays the supported languages written natively.

If user selects 'हिंदी', then the translated page is shown and 'हिंदी' is displayed on the top right. The language stops changing. The drop-down menu could still be available.

If the user scrolls down without selecting a language, it defaults to 'English'.  Again, the language stops changing and drop-down menu is still available.

However, I can see this become distracting if this happened at the top of every page. Maybe once a language is selected, all other pages stop doing the changing animation.

But I'm also fine with Joe's latest suggestion, because it's simpler. I agree on using 'English' rather than 'Language' for consistency. So whatever language the site is on, the drop down menu options are always 'English', 'Nederlands' and 'हिंदी'.

rgommers:
> Most examples I'm seeing display the selected language, so we would render 'English' instead of 'Languages' at the top level. Inside the dropdown would be supported languages written natively.

+1 for this. I lived through the Flash and Dreamweaver era - anything that moves is bad.

shaloo:
+1 and pls keep it simple.  For e.g., see Uber - they keep language in footer and do not auto-select based on browser language / location unless a user chooses to change it.  I like their language selection method too - simple, clear and obvious.

joelachance:
Kept the language dropdown where it currently is, and removed the JS. Please re-review!

rgommers:
LGTM, merging. Thanks @joelachance!

PR #161, opened 2020-01-29
Recover script to create the EHT imaging dependency graph

rgommers:
This is for the black hole case study. Created by Shaloo in gh-23,
it got lost when we renamed `master` to `old-sphinx-site`.


rgommers:
Thanks for pointing that out @shaloo

PR #162, opened 2020-01-30
Feature/156 content layout improvements

### LINKED ISSUE #156: Review content pages
Let's improve the content page styling a bit.

- [x] The current width is too wide for desktop. I think a `max-width: 900px` seems reasonable.
- [x] Some of the shortcuts are broken. Make sure we're using `h2` and `h3` headers.
- [x] Update shortcuts width to make it slightly wider to avoid text wrapping.

Create any necessary shortcodes to improve content pages.

joelachance:
Resolves #156


static/js/app.js
```diff
@@ -84,13 +84,24 @@ const shortcutsTarget = $('#shortcuts');
...
+        scrollTop: $(`#${elTitle}`).offset().top-60
     }, 1000);
     })
   });
+}
+
+// Removes the shortcuts container if no shortcuts exist.
+// Also removes the 'Get Help' link.
+if ($('#shortcuts div').length < 1) {
+  $('.shortcuts-container').css('display', 'none');
```
joelachance: I'm not sure this is wanted. Curious everyone's thoughts here. Another option might be to simply remove the 'Shortcuts' title, however, i think the `Get Help?` link might want to live somewhere else so it's not the only element on the left sidebar.

rgommers: I agree, it's kind of weird to have the sidebar with a single link. This looks better. Let's merge and deal with the get help separately - it could live in the footer for example

Issue #57, opened 2019-10-18
Cross platform and browser testing for Hugo generated pages


InessaPawson:
The website should work entirely in the last few versions of the most popular platforms (macOS, Linux, Windows, Android, iOS) and browsers (Chrome, Opera, Firefox, IE/Edge, Safari), and provide an acceptable experience on IE 8 and 9. It must be WCAG AA compliant.

MarsBarLee:
Could I take a stab at this? I'll incorporate the [test site Joe suggested](https://www.lambdatest.com/). I don't have much experience making something WCAG compliant besides alt text for images but I can look into that as well.

joelachance:
Of course! Curious how it goes/what you find.

MarsBarLee:
Quick update: most issues I've seen so far are related to screen/device size, rather than operating system or browser. In many mobile devices the footer is missing, the navigation bar icon is repeating twice and clipping itself or too much white space. I'll look into CSS media queries and start on that tomorrow.

Issue #168, opened 2020-02-20
Rendering seems to be off on high resolution screens


seberg:
On my computer, the website is rendered not very well. Also at the bottom the case studies are not all three next to each other. This happens both with chromium and with firefox.

See attached image, let me know if you need some other information.
![Screenshot from 2020-02-19 18-45-08](https://user-images.githubusercontent.com/61977/74896099-39c23100-5348-11ea-9c0a-e19bbee6a741.png)

joelachance:
@seberg, could you be more specific? We're aware of the whitespace and are going to re-style very soon.

We're going to comb over the homepage and fix all the spacing and minor issues, which I think you might be alluding to.

seberg:
Hmm, yeah, maybe the abundance of whitespace is just more extreme, the case studies are not all next to eac hother as well, but not sure if that is not the same, known, problem:
![Screenshot from 2020-02-20 08-12-59](https://user-images.githubusercontent.com/61977/74955141-0c619b80-53b9-11ea-961d-cf5e2f773674.png)

joelachance:
Typically it's good to not let content spill out too wide. I can tell your resolution is pretty high too, which will add to the whitespace a bit :) That being said, I agree, we can get those all in one row.

seberg:
I am not actually what bit me/us the wrong way, maybe the font also look just small compared to the space the whole layout takes?

joelachance:
Good observation, I'll take a look!

MarsBarLee:
This is how the site looks like in 2560 x 1440, which was the resolution the issue opener used.
![numpyoriginal](https://user-images.githubusercontent.com/46167686/76905551-8d0c9e00-6878-11ea-8904-51b43da64759.png)

This is the mock-up with some changes I'm thinking of implementing.
![numpyedited](https://user-images.githubusercontent.com/46167686/76905577-9ac22380-6878-11ea-89d6-0f26eec110a0.png)
Changes include centering the content to the middle of the page and removing the white space.  

Does anyone have any thoughts or suggestions?

joelachance:
I'm updating the shell (above the fold) section, so expect that to change. I'm moving the NumPy title/CTA & the shell into rows vs. columns, so I think it'll look nicer and more centered.

As far as the rest, I think what you have looks nice! I'd like to try and maintain a max-width similar to the width of the nav bar, which I believe matches what you have in your layout.

seberg:
[...]

Issue #118, opened 2019-12-23
Website Accessibility


joelachance:
We need to check that the new website can be navigated with browser accessibility tools.

- [x] Test happy paths and make sure there aren't any bugs when navigating with accessibility tools.

MarsBarLee:
As suggested in a previous PR involving web-accessibility, I'll look into adding aria-labels.

joelachance:
Some additional changes on branch: feature/118-accessibility

joelachance:
@MarsBarLee @rgommers @InessaPawson 

I just did a quick once-over with Mac's VoiceOver and things look pretty good. CodeMirror (our interactive shell) isn't set up very well for accessibility, but I'm not sure there's much we can do there. Overall, I was able to navigate the site without any major issues.

One callout is that the install instructions aren't called out as code blocks in VoiceOver, but since those are being rendered using some Hugo magic, I'm not sure we can tag those with aria labels.

InessaPawson:
This is above and beyond, @joelachance! Thank you for doing this!

PR #170, opened 2020-02-24
Cross platform and browser testing for Hugo generated pages and WCAG AA compliance

MarsBarLee:
Fixes gh-57

# Changes made:
### WCAG
- Added alternative text for all images and logo. 
- Made alternative text more specific with the type of graphs made in the Ecosystem section . For example, 'vispy image' -> 'A Voronoi diagram made in vispy'.

### Mobile and tablet view
- Centered columns and maintain image scale in 'Case Studies' section.

### Cross platform testing
- Most problems so far about how the site looks in mobile and tablet view, which falls under cross platform testing for iOS and Android.

### Cross browser testing
- Need to do more testing.

# To Do:
- Will do more for WCAG 'AA' level-specific compliance, such as the checking and changing the contrast ratio.
- More cross platform testing for macOS and Linux.

# Question
- Would it be better to separate this PR into a few different issues, such as one for cross-platform testing, one for browser testing and one for WCAG compliance?


joelachance:
@MarsBarLee to answer your question, we do have an accessibility ticket already which I tagged you in #118. My thoughts there were that we could test our site using some popular screen readers (Apple Voice Over, JAWS, etc.) and see how our site does. I have a feeling we may need to add some aria labels. I think alt text is fine in this PR, but we may want to move anything else accessibility-related into a new PR.

How are you testing the site for mobile? I've always used Chrome's 'Device Toolbar' in DevTools, which I think does a pretty decent job. I've also been trying to make sure the page is responsive for all widths on desktop as well.  I did find some text hidden in section 3 when adjusting the page width on desktop. Also, I'm not sure we want all text to have the link underline on rollover. I do like the `cursor: pointer` however!

LGTM, feel free to merge once you feel this is no longer WIP.

MarsBarLee:
Hey @joelachance , thanks for the comment! I missed the gh-118, my bad. I talked to Ralf about this PR and I'll remove the WIP and ask to merge the changes I made so far. Then start working on the more specific PRs, like gh-118 and gh 168, which were recently assigned to me. 

I'll look into aria-labels for gh-118.

If the extra text on section 3 is in wide desktop view, I'll see if I can fix it in gh-168.

I have been using mostly Chrome DevTools, for different desktop sizes and mobile views. I've been also using LamdaTest, but it's been giving me some issues that aren't present in Chrome DevTools, such as missing footers in Lambda Test but not in Chrome DevTools.

For the link underline, would that small fix be in a separate PR?

joelachance:
@MarsBarLee, no worries if you had missed that ticket, there's a lot floating around :)
Good find with LambdaTest, I didn't know about that one!

I think we can either include the link underline in this PR or in another one, it should be a very small change. Up to you!

MarsBarLee:
@joelachance I decided to remove the underline in this PR, since it's a small change like you mentioned. :)

static/css/section3.css
```diff
@@ -1,3 +1,7 @@
+a:hover {
+    text-decoration: none !important;
+}
+
```
joelachance: I'm not sure we quite want this-- all anchor elements won't have text-decoration at all with this rule. I think we might want this for actual URL's, but just not for this specific instance. I think this might work, but need to try it out:

```
.section-3 a:hover {
    text-decoration: none;
}
```

et HEAD~1wq
```diff
@@ -0,0 +1,2467 @@
+[33mcommit b41edfd2243208c90e18e6a3d3f91881bfd7f5bb[m[33m ([m[1;36mHEAD -> [m[1;32mmaster[m[33m)[m
```
rgommers: Please delete this file, looks like you committed it by accident @MarsBarLee

rgommers:
One of the `section3.css` size changes causes the box sizing to be messed up:

image

I've tried with both Firefox and Chrome, with some full rebuilds - should be real.

MarsBarLee:
I added more media queries for 'middle' screen sizes, that are bigger than tablet portrait view but smaller my monitor size. I ran simulations on mine, but could you check again if they look good on yours?

rgommers:
It looks to me like the width of the tiles themselves needs to be a bit more restricted or the aspect ratio constrained, they can get super wide now (_disclaimer: I know little about best practices for this_). Here's some test results:

My regular browser size:

image

At ~75% width:

image

At ~150% width:

image

rgommers:
Actually the behavior in my comment above is what happens on master as well, so at least the regression is fixed by your last commit @MarsBarLee. So perhaps we can merge as is, and deal with the rest later. All the other fixes in this PR would be nice to get in. @joelachance does that sound okay?

joelachance:
Yep! Looks good to me.

rgommers:
One closer to the finish line:) Thanks @MarsBarLee and @joelachance

PR #176, opened 2020-03-14
Remove Hindi translations.

rgommers:
We decided already those were not needed (just added for testing),
so removing now to reduce content and make searching the repo easier.
Content wasn't right anyway, it was just English.

Leaving the Dutch translation for now (one language left for testing),
that will be replaced with Mandarin soon.


rgommers:
All green, merging.

Issue #102, opened 2019-12-11
Installation instructions can be confusing to newcomers


stefanv:
The current installation instructions are beautifully simple; well done!

That said, they may be confusing to newcomers.  Conda: "To install NumPy do X or Y" without giving advice on when to use which.  When using pip do X or Y if Python comes with your OS.  Most users may not have any idea whether Python comes with their OS, or whether that is the version they're using.  Perhaps `pip install numpy --user` works fine in a venv too, and can be the only command recommended.

rgommers:
Yep, still going to polish this. Also, we want to have a widget like on https://pytorch.org/get-started/locally/, @InessaPawson is working on that.


>  Perhaps `pip install numpy --user` works fine in a venv too, and can be the only command recommended.

Nope, `--user` with virtual/conda envs is terrible imho, since it _couples_ environments that should be separate from each other. The only reason to use `--user` is in a Linux distro style setup, where Python is installed in a place with restricted permissions (e.g. with `apt`). It's a thing from the past I think, using environments is the thing to teach these days.

matthew-brett:
`pip install --user` does not work in a virtualenv, sadly:

```
$ pip install --user transforms3d
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
```

I personally run with my --user environment nearly all the time, and dip into virtualenvs when I have to check dependencies or try another Python version.  That's what I'd recommend for pip users.

rgommers:
> `ERROR: Can not perform a '--user' install.`

nice, that's an improvement over what `pip` used to do, which was just install anyway and create a mess.

Unfortunately there's no universal command that can be given. I think we have to have concise install instructions (as we have now), but add some notes for new users that link out to a page with more detail, to explain about distributions, ways to obtain Python, etc.

I think we should really be teaching use of per-project/activity environments (whether conda or virtualenv). Anything else tends to end in:

![image](https://user-images.githubusercontent.com/98330/70669415-32392a00-1c2b-11ea-9f75-13c971907cb5.png)

matthew-brett:
I think having multiple environments adds a layer of complexity to the install that will only be useful for a relatively small subset of people (i.e. developers, working on more than one project).

[...]

PR #179, opened 2020-03-14
Add a full install guide

rgommers:
This simplifies the main install page further (as requested in gh-102) and adds a much larger guide in a second page, with recommendations based on user experience level and OS, as well as some important background on pip/conda, on reproducible installs, and on MKL/OpenBLAS depending on install method.

Closes gh-102. There's a separate issue for the install widget, that won't happen before the public launch of the website.


rgommers:
In case no one gets around to reviewing soon: I plan to merge this before the website launch (planned tomorrow evening, but likely a couple of days late). We can still review and iterate after.

mattip:
Should this mention activating the conda environment before running python scripts?

rgommers:
@mattip I don't think so, that's very specific and is anyway recommended in the regular `conda` docs (it's hard to avoid even, expect in some VSCode workflow). That said, I did think about including typical workflow with conda and with pip. Something like:
```
# Step 0: install Miniconda
conda create -n python=3.7 someproject
conda activate someproject
conda install numpy scipy matplotlib scikit-learn  # put this in environment.yml later
python analysis_script.py
```

```
# Step 0: install Python from, e.g., python.org
python -m ensurepip  # install Pip
# This puts everything in a single environment - use requirements.txt for this later
pip install numpy scipy matplotlib scikit-learn
python analysis_script.py
```

```
Step 0: install Python from, e.g., python.org
Step 1: install Poetry (https://python-poetry.org/docs/#installation)
poetry new someproject  # or add/edit pyproject.toml for an existing project
poetry env use python3.7
poetry add numpy scipy matplotlib scikit-learn
python analysis_script.py

**Comment by rgommers** [comment]
Some other thoughts:
- Ideally a lot of this content will live on a new SciPy/PyData-wide docs site at some point, the "entry portal to the ecosystem". Since there's no good place to refer people to now, I think putting it here is much better than not having this content.
- Writing install instructions for pip & co is still really annoying, there's no one way to recommend and there's a whole zoo of tools.
- Went with Poetry because it's much better maintained than Pipenv at this point.
- The content in the [Python Packaging User Guide](https://packaging.python.org/) isn't useful to refer to, outdated, mixes dev and user info, and doesn't take the SciPy/PyData user (now >50% of all Python users) into account at all.

**Comment by rgommers** [comment]
@takluyver you usually have interest and good insights in this stuff as well - any feedback is very welcome.

File: content/en/install.md [hunk 1]
```diff
@@ -15,22 +15,11 @@ If you use `conda`, you can install it with:
...
 ```bash
 pip install numpy
 ```
-
-If your Python came with your operating system, we advise you either install a separate Python or use:
-
-```bash
-pip install numpy --user
-```
+If your Python is installed in a place where you don't have user-level write permissions (e.g. because it came with your operating system), we advise you either install a separate Python or use `pip install numpy --user`.
```
takluyver: As of pip 20.0 (released a couple of months ago), it will automatically detect when it needs to do a user installation based on permissions, so it shouldn't be necessary to mention this.

Of course, you may want to keep it for a while longer in case people still have older versions of pip, but as pip warns when it's outdated, hopefully most people will be reasonably up to date.

content/en/install.md
```diff
@@ -15,22 +15,11 @@ If you use `conda`, you can install it with:
 conda install numpy
 ```
 
-or with:
-
-```bash
-conda install -c conda-forge numpy
-```
-
 ## pip
```
takluyver: I might list pip above conda, because I figure that most people with conda either already have numpy (installed Anaconda and using the default environment) or know enough to get it anyway. But I'm sure you can make a case the other way too.

content/en/install.md
rgommers: Good point, I'll remove it. The case where `pip` can't be easily upgraded (Linux distros), numpy should anyway be installed with the distro package manager.

content/en/install.md
rgommers: I was thinking:
- if people have `pip` but not `conda`, they either know to read on, or if they think "what's conda" then it's in their benefit to learn. Too many resources like the Python Packaging User Guide are not explaining conda at all.
- if people have `conda`, they should use it - we've seen a _lot_ of cases of people using `pip install numpy` within conda.

rgommers:
No more comments, so merging. Thanks again for reviewing @takluyver

PR #184, opened 2020-03-18
Feature/152 clean up

### LINKED ISSUE #152: Fix Homepage spacing
There's too much whitespace between content on the homepage. Improve spacing between sections on the homepage.

- [x] Give the shell 4 borders
- [x] Shrink the space above and below the shell
- [x] Fix key features media query (Goes back to a single row for very wide desktops)
- [x] Change key features background to separate sections
- [x] Remove any additional spacing around sections with too much

joelachance:
This refactors the shell into a row layout from a column layout.
I've updated the shell colors and refactored the lesson content layout (we still need this!) into a partial for easier editing.

As previously noted, the shell is finicky, and this is mainly due to thebelab integrations. I don't feel we have any blockers at this time, but some things we can fix post-launch. The main issue i've found is once the shell is focused, it then inserts the `import numpy as np`, and the clickable area isn't quite where you'd expect. We could move the caret slightly to change the expectation here, if people think that is a good short-term fix.

partially fixes #152


joelachance:
Does anyone know anything about this CI build failure?

rgommers:
Nice @joelachance:) I like the terminal color. Did notice the cursor disappears under the Run button:

image

Also the "while we wait" text has a backslash and link (should be https://numpy.org/doc/) isn't there yet:

image

That `import numpy as np` appears when you click on the terminal when it's active is a nice touch. Could add a line break, so the cursor starts on the next line, rather than after `np`. Or 2 line breaks - PEP8 and all:)

rgommers:
You could may trim the white space even more above/below the first section with the logo and get started button. This is what it looks like on my 32" screen:

image

On a smaller screen the Key Features won't be visible at all. Spacing like between the bottom of the terminal section and Key Features should be enough I think.

rgommers:
> Does anyone know anything about this CI build failure?

Seems to have fixed itself. Hasn't been reliable unfortunately.

rgommers:
> the lesson content layout (we still need this!)

Yes, it's on my radar. There's a few more content things to fix.

MarsBarLee:
On smaller screens, the shell is overlapping into the next section. This is how it looks on my 1280 x 720 screen.
numpy small screen

When I try out other screen resolutions, such as 2560 x 1440, the site looks the same as Ralf's version on 32" screen

joelachance:
@rgommers 

> Nice @joelachance:) I like the terminal color. Did notice the cursor disappears under the Run button:

There's not a great way to update this since CodeMirror will let you have unlimited lines. I added some max-height to this div to keep it from going that low, but that's unfortunately a shortcoming of this integration.

joelachance:
@rgommers 

> Also the "while we wait" text has a backslash and link (should be https://numpy.org/doc/) isn't there yet:

 Updated!

joelachance:
@rgommers 

> That `import numpy as np` appears when you click on the terminal when it's active is a nice touch. Could add a line break, so the cursor starts on the next line, rather than after `np`. Or 2 line breaks - PEP8 and all:)

Updated this so it's a little better! Give it a shot and let me know what you think.

I did attempt previously to add better spacing, but I believe there's some pretty heavy JS going on in the CodeMirror logic, and it's pretty hacky to get it to where we would want it :/

joelachance:
@MarsBarLee 

> On smaller screens, the shell is overlapping into the next section. This is how it looks on my 1280 x 720 screen.

Fixed! Good find.

joelachance:
@MarsBarLee @rgommers 

Please re-review when you have a moment. Thank you!

rgommers:
The section headers (Key Features etc.) are now left-aligned instead of centered. Is that on purpose? Centered looked better I think

image

Also did you intend to change the background of the tiles to gray?

Case study tiles scaling for varying browser width is better now, looks good.

> Updated this so it's a little better! Give it a shot and let me know what you think.
>
> I did attempt previously to add better spacing, but I believe there's some pretty heavy JS going on in the CodeMirror logic, and it's pretty hacky to get it to where we would want it :/

Looks like only 7 lines fit above the button. Typing further lines is possible, but they are now completely hidden. Everything below the button is black or for output. Also, the background of the output is white now, I believe you had it as transparent before:

<details>

image

image

</details>

My impression is that the 7 lines above is okay, the overflow is hard to deal with so maybe leave as is, the white background works sometimes - here it looks nice actuallly (in contrast to what's in "details" above:

image

joelachance:
@rgommers 

> The section headers (Key Features etc.) are now left-aligned instead of centered. Is that on purpose? Centered looked better I think

This was on purpose, but I can center it.

> Also did you intend to change the background of the tiles to gray?

Yes, but I can change this back too if you'd like.

> Looks like only 7 lines fit above the button. Typing further lines is possible, but they are now completely hidden.

Correct. There's no way to limit the amount of new lines that I know of. I think this is all controlled via JS in CodeMirror logic.

> Also, the background of the output is white now, I believe you had it as transparent before

This was always white, but I can try and make it transparent! The caveat being that we'll also need to make the text white, and any plotting might not be completely visible.

I can try and deal with the overflow, but it's hard to avoid __all__ bugs since this is a fully working Juypter notebook, and anyone could put anything in here.

Please advise.

joelachance:
@rgommers @MarsBarLee any thoughts on the 'Learn More' button in Key Features? I thought more 'call to action's might be nice.

rgommers:
> @rgommers @MarsBarLee any thoughts on the 'Learn More' button in Key Features? I thought more 'call to action's might be nice.

I like this, nice touch



> This was always white, but I can try and make it transparent! The caveat being that we'll also need to make the text white, and any plotting might not be completely visible.
> 
> I can try and deal with the overflow, but it's hard to avoid **all** bugs since this is a fully working Juypter notebook, and anyone could put anything in here.
> 
> Please advise.

Agreed, it's hard. Let's leave as is I'd say. The white is nice except when it overflows, and then it becomes messy anyway.



> > The section headers (Key Features etc.) are now left-aligned instead of centered. Is that on purpose? Centered looked better I think
> 
> This was on purpose, but I can center it.
> 
> > Also did you intend to change the background of the tiles to gray?
> 
> Yes, but I can change this back too if you'd like.

It's a taste question - personally I'd prefer to change both of those back. The white tiles looked very clean and modern, and everything else is also centered so centering titles makes sense to me.

joelachance:
> Agreed, it's hard. Let's leave as is I'd say. The white is nice except when it overflows, and then it becomes messy anyway.

@rgommers I shrunk the top section with the logo and added some height to the shell. I also added an overflow rule to keep things in bounds, so I think this is a little better! It's still possible to get things not where we want them, but it's a touch harder to 'break'.

Centered the headers and changed the color back to white, let me know what you think!

rgommers:
I think this looks good, thanks Joe! The terminal has a scroll bar now for the output, that looks nice.

Let's get this in!

One potential other tweak is to move the separator between input and output, plus the "Enable / Run" button, down a little - the input area for writing code is still 7 lines, while the output area is much larger and can scroll if needed.

InessaPawson:
>...personally I'd prefer to change both of those back. The white tiles looked very clean and modern, and everything else is also centered so centering titles makes sense to me.

I second @rgommers on the above. 

And how about "Get Started" button on the homepage?  Are we linking it to anything? (e.g. /install)

InessaPawson:
@joelachance I like the enhanced interactivity of the shell. Good job!

rgommers:
> And how about "Get Started" button on the homepage? Are we linking it to anything? (e.g. /install)

Fix in gh-189

PR #189, opened 2020-03-22
Make "Get started" button link work

rgommers:
We have no dedicated "Get started" page. I looked at some other projects, some link to install procedures, some to a quickstart or similar. For now I thought linking to the `Learn` page is most appropriate. That has the (not so great) quickstart guide in the Sphinx docs, an install link, and a bunch of other resources.

### DIFF
--- a/layouts/partials/hero-body.html
+++ b/layouts/partials/hero-body.html
@@ -18,7 +18,7 @@
                 </div>
                 {{ end }}
                 <div class="hero-cta">
-                    <button class="cta-button">{{ $buttonText }}</button>
+                    <a href="{{ $buttonLink }}"><button class="cta-button">{{ $buttonText }}</button></a>
                 </div>
             </div>
         </div>
--- a/config.yaml
+++ b/config.yaml
@@ -21,7 +21,7 @@ params:
     # Button text
     buttontext: Get started
     # Where the main hero button links to
-    buttonlink: "#"
+    buttonlink: "/learn"
     # Hero image (from static/images/___)
     image: logos/numpy.svg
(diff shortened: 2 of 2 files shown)


layouts/partials/hero-body.html
```diff
@@ -17,8 +17,8 @@
                     {{ . }}
                 </div>
                 {{ end }}
-                <div class="hero-cta">
-                    <button class="cta-button">{{ $buttonText }}</button>
+                <div class="hero-cta" href="/learn">
```
joelachance: the href attribute on the div here won't do anything, but the code below (the anchor wrapping the button) looks great!

rgommers: Good catch, that was a leftover from trying to figure out what worked. Fixed now.

InessaPawson:
@rgommers Makes sense to me. Do you think we need a dedicated "Get Started" page?

rgommers:
I think a rewrite of the quickstart page (already planned) would be the way to go, rather than a new page.

shaloo:
I can take that up next once I am done cleaning up 45 and case studies.  Is that ok?

rgommers:
> I can take GettingStarted/Quickstart up next once I am done cleaning up 45 and case studies. Is that ok?

@shaloo I think it will be easier to rework https://numpy.org/devdocs/user/quickstart.html after the website is launched, because then it can link back to Install and Learn. What you can do now though is to review it and make an overview of what needs changing. We've recently started doing this (see https://github.com/numpy/numpy/issues/15760 for an example), planning out what should be on a docs page and reviewing that should yield better results than just starting to write.

shaloo:
Adding this work-item to post launch to-do list along with #213

PR #190, opened 2020-03-22
Update Fresh theme to latest. This gets rid of two build warnings

rgommers:
Warnings were (coming from the sidebar template in the theme):
```
WARN 2020/03/22 17:11:18 Page.Dir is deprecated and will be removed in a future release. Use .File.Dir
WARN 2020/03/22 17:11:18 Page.URL is deprecated and will be removed in a future release. Use .Permalink or .RelPermalink. If what you want is the front matter URL value, use .Params.url
```

I've done some testing and looked at the other commits in the 3 months that we're forwarding the submodule for, no obvious issues.


joelachance:
LGTM, seemed to load just fine after a `git submodule update --init`! Merging.

PR #192, opened 2020-03-22
Fixes hugo loading error with updated theme

joelachance:
Quick fix for the hugo-fresh theme update. I thought this might be part of not running the submodule init, but I think this is a bug.


rgommers:
Thanks Joe. I think it's indeed better to quote - I could reproduce only on one of my machines, depends on Hugo version.

rgommers:
Hugo doesn't seem so stable unfortunately.

joelachance:
Agreed :(

Issue #81, opened 2019-11-24
Create copy for the NumPy terminal


joelachance:
We need to create copy above the NumPy terminal on the homepage. A welcoming message and perhaps a code sample might be nice. Ideas welcome.

rgommers:
(moving comment from @shaloo to here):

Here are some more ideas for copy / layout and example code in terminal on the landing page - refer to attached image, blue text.

<img width="1247" alt="numpy-issue83-inputs" src="https://user-images.githubusercontent.com/5890484/69939774-a9254300-1506-11ea-9fb9-4a017f788d24.png">

stefanv:
Probably worth getting PEP8 right on the landing page, so `np.linspace(0, 10)` instead of `np.linspace(0,10)`, etc.

rgommers:
I've thought about it some more; it's quite hard to show much in a few lines of code. 

Perhaps something like this, showing array creation, indexing and using a method over one axis:
```
>>> x = np.arange(15, dtype=np.float64).reshape(3, 5)                              
>>> x[1:, ::2] = -99                                                               
>>> x                                                                              
array([[  0.,   1.,   2.,   3.,   4.],
       [-99.,   6., -99.,   8., -99.],
       [-99.,  11., -99.,  13., -99.]])
>>> x.max(axis=1)                                                                  
array([ 4.,  8., 13.])
```

stefanv:
Maybe make the dtype `int64` to save on all the printing of dots?

rgommers:
Okay:)
```
>>> x = np.arange(15, dtype=np.int64).reshape(3, 5)                                
>>> x[1:, ::2] = -99                                                               
>>> x                                                                              
array([[  0,   1,   2,   3,   4],
       [-99,   6, -99,   8, -99],
       [-99,  11, -99,  13, -99]])
>>> x.max(axis=1)                                                                  
array([ 4,  8, 13])
```

PR #194, opened 2020-03-22
Add content for the lesson displayed when the shell is active.

rgommers:
Note that the styling for this needs fixing in `shell.css`, now nothing is visible. I tried with the `shell-content-message` class and got:

image

So the highlighting is fine, but the text justification is messed up. @joelachance could you help with that?

Closes gh-81

### DIFF
--- a/layouts/partials/shell-content.html
+++ b/layouts/partials/shell-content.html
@@ -9,5 +9,5 @@
 </div>
 
 <div class="shell-lesson shell-content" style="display: none;">
-    Here we should include a lesson about numpy!
+    {{partial "shell-lesson.html" | print | markdownify}}
 </div>
--- a/layouts/partials/shell-lesson.html
+++ b/layouts/partials/shell-lesson.html
@@ -1,0 +1,26 @@
+
+The standard way to import NumPy:
+
+```python
+>>> import numpy as np
+```
+
+Create a 2-D array, set every second element in some rows, and find max per
+row:
+
+```python
+>>> x = np.arange(15, dtype=np.int64).reshape(3, 5)
+>>> x[1:, ::2] = -99
+>>> x
+array([[  0,   1,   2,   3,   4],
+       [-99,   6, -99,   8, -99],
+       [-99,  11, -99,  13, -99]])
+>>> x.max(axis=1)
+array([ 4,  8, 13])
+```
+
+Generate some random numbers with an exponential distribution:
+```python
+>>> rng = np.random.default_rng()
+>>> samples = rng.exponential(size=2500)
+```
--- a/static/css/shell.css
+++ b/static/css/shell.css
@@ -84,10 +84,18 @@
 
 .shell-lesson {
     display: none;
-    align-items: center;
-    justify-content: center;
-    height: 30%;
-    width: 100%;
+    text-align: left;
+    padding: 15px;
+    height: 100%;
+    width: 60%;
+}
+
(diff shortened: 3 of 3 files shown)


joelachance:
It might be easiest to share changes here:

```
.shell-lesson {
    display: none;
    text-align: left;
    padding: 15px;
    height: 100%;
    width: 75%;
}
```

Let me know if that looks better to you, @rgommers!

rgommers:
Thanks. That's a good start, however the padding/margin inside that box is huge so it expands the size of the whole section:

image

From my clumsy use of the inspector, I think it's mainly the padding and margin of the `<p>` elements within the `shell-lesson` area.

rgommers:
And the flex elements in between `<p>` and `<code>` blocks

rgommers:
Hmm, I don't see how to override margin/padding of child elements. Perhaps `markdownify`'ing the separate code blocks and putting the comments inside `shell-content.html` so they're easier to control makes sense?

joelachance:
Try this rule:

```
.shell-lesson > pre {
    margin: 10px 0 5px 0;
}
```

That will definitely shrink margin between those blocks and we shouldn't need to refactor content outside of the partial you've created.

rgommers:
Thanks @joelachance, getting there. I still don't understand these precedence rules. Works for `<p>`:

image

Does not work for `<div class=highlight>` or `<pre>` inside that:

image

 
image

With this CSS (tried a bunch of variations:
```
.shell-lesson {
    display: none;
    text-align: left;
    padding: 15px;
    height: 100%;
    width: 60%;
}

.shell-lesson > p {
    margin: 2px 0 2px 0;
}

.shell-lesson > pre {
    margin: 5px 0 0 0;
}

.shell-lesson > div {
    margin-bottom: 5px;
}
```

rgommers:
If I toggle `margin-bottom: 1.5rem;` off in the inspector, things look good.

joelachance:
@rgommers 
We need to create a more specific rule in order to 'override' what's already in there. The rules that aren't working simply aren't specific enough.

In the first example, `.shell-lesson > p` is more specific than uikit's rule to assign `margin-top` for all `p` tags. UIKit is assuming you want to use them for the entire site which is why they make global rules like this. Unfortunately, hugo-fresh is doing exactly the same (any styling inside a .sass file), so we're having to combat all of those rules.

The `.shell-lesson > p` rule is saying that all `p` elements that are descendants of `.shell-lesson` should have the assigned rules. Instead of the `>`, a space will also work. This is a more specific rule, thus the rules we assign it will propagate. This post explains it: https://techbrij.com/css-selector-adjacent-child-sibling

The `.highlight:not(:last-child)` rule is already pretty specific (I'm not able to find where this `.highlight` class lives inside the shell-lesson?), which is why it's not working. It's not exactly best practice, but you can use a rule like this:
```
.highlight:not(:last-child) {
margin-top: 1px !important;
}
```
You'll notice we've done it a couple of places already in our stylesheets, which I think is sometimes better than having to assign an element more id's or classes just to avoid the `!important`.

If you can point out where the `.highlight` classes are we'd like to update the margin for, I can PR those changes.

rgommers:
Cool, thanks for the explanation! I'll try.

`highlight` comes from the Chroma highlighter, which is invoked by `markdownify`, I believe.

rgommers:
Okay that worked! This is good to go I think:

image

PR #195, opened 2020-03-22
Improve the content and layout of the "Key Features" tiles.

rgommers:
Change to six tiles (works better for some screen widths, not 3 tiles on row 1 and 1 tile on row 2) and trim the text on each tile. 

image


joelachance:
oops, sry @stefanv! feel free to review, and I can update.

PR #197, opened 2020-03-24
Web accessibility - Contrast ratio for WCGA AA

MarsBarLee:
Increased contrast ratio between words and background by changing the color of links in the Key Features, Ecosystem sections and footer.

This meets the contrast ratio a WCGA AA rating. Web Content Accessibility Guidelines and its rating range from A to AAA.

I used WebAIM and WAVE tools to check the rating.

**Questions**
I used ! important in the internal css to override external linked css from UiKit. I know ! important isn't generally recommended but it's how I was able to override the external css. **Does anyone know another way or is using ! important here OK?**

**Can I remove the empty h3 tag in footer.html and empty title in config.yaml?** Or is there a title planned for these sections?
image

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -102,7 +102,6 @@ languages:
             - text: Roadmap
               link: https://numpy.org/neps/roadmap.html
         column2:
-          title: ""
           links:
             - text: About us
               link: /about
@@ -115,7 +114,6 @@ languages:
             - text: Contribute
               link: /contribute
         column3:
-          title: ""
           links:
             - text: Terms of use
               link: /terms
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -12,7 +12,7 @@
         <div class="link-column">
           <div class="footer-column">
             <div class="footer-header">
-                <h3>{{ .title }}</h3>
+                <!-- <h3>{{ .title }}</h3> -->
             </div>
             <ul class="link-list">
               {{- range .links }}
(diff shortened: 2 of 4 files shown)


MarsBarLee:
I'm looking into the error 'sudo: hugo: command not found' in the Travis CI. 

When I do git status, it tells me that 'modified:   themes/hugo-fresh (modified content)'. I didn't commit that change since I don't think I made any changes in the theme or wanted to.

Could the modified theme be related to the CI error I'm getting?

joelachance:
> Does anyone know another way or is using ! important here OK?

If you can write a more specific CSS rule, you won't need `!important`. I've used it a couple of times just because sometimes those global rules are tough to circumvent (and I'm a bit lazy ;)) 

> Can I remove the empty h3 tag in footer.html and empty title in config.yaml?

If it's not being used, sure!

> Could the modified theme be related to the CI error I'm getting?
@rgommers? Any thoughts here?

MarsBarLee:
@joelachance Actually, I think I figured how to override without using ! important last night, and the commits here don't use ! important. Solved it already, haha. I also pushed another commit to remove the headings.

rgommers:
> > Could the modified theme be related to the CI error I'm getting?
> 
> @rgommers? Any thoughts here?

It's flaky like crazy, just ignore it for now. I've seen `sudo: hugo: command not found` on a number of other PRs, it's not related the changes in this PR.

MarsBarLee:
@rgommers Ok! Do I wait for the build to work later then?

rgommers:
Just ignore it, if you think the PR is ready then we can review and merge

PR #200, opened 2020-03-25
Fix the link to NumPy documentation

InessaPawson:
Fixing the link to NumPy documentation. The correct link is https://numpy.org/doc/


rgommers:
Thanks @InessaPawson. I agree with this change. The other link would work as well once deployed, but this is more robust and works for local dev builds as well. No need to keep the relative link I think, given that they're really two separate sites that happen to share a base URL.

PR #201, opened 2020-03-25
Minor edit in the footer

InessaPawson:
Minor edit in the footer.


rgommers:
Thanks @InessaPawson. This change makes sense. Current master on a 13" screen looks like:

image

So wrapping after `resources,` makes this look better.

PR #202, opened 2020-03-26
Add a link to https://numfocus.org

InessaPawson:
Since we are inviting to visit the website, let's link to it as well.


rgommers:
Good catch, thanks Inessa

Issue #45, opened 2019-09-21
Idea #1 for Team page


rgommers:
This is generated from a simple yaml file that then fetches data from GitHub and creates a gallery: https://datapythonista.github.io/pandas-web/community/team.html

Code at https://github.com/datapythonista/pandas-web/blob/master/pysuerga.py#L43
Thanks for the suggestion, @datapythonista!

datapythonista:
The code has now been committed to pandas, the new links are:
- https://dev.pandas.io/community/team.html
- https://github.com/pandas-dev/pandas/blob/master/web/pandas_web.py

If you can be interested in the whole `pandas_web.py` script, I'm happy to move it to a separate repo and create a conda package for it. Probably after we integrate the new website with Sphinx. Since we'll need to render the naviagation bar from the yaml in Sphinx too.

InessaPawson:
Very nice, @datapythonista! 
The new active link is dev.pandas.io/about/team.html

shaloo:
> This is generated from a simple yaml file that then fetches data from GitHub nad creates a gallery: https://datapythonista.github.io/pandas-web/community/team.html
> 
> Code at https://github.com/datapythonista/pandas-web/blob/master/pysuerga.py#L43
> Thanks for the suggestion @datapythonista

Hi, the first link team.html is not working.  @datapythonista, @rgommers is this information available somewhere else?  I checked datapythonista.me domain as well but couldn't locate it.

datapythonista:
Sorry @shaloo, those pages were deployed, and I removed my temporary copy.

- https://datapythonista.github.io/pandas-web/community/team.html -> https://pandas.io/about/team.html
- https://github.com/datapythonista/pandas-web/blob/master/pysuerga.py -> https://github.com/pandas-dev/pandas/blob/master/web/pandas_web.py

shaloo:
@datapythonista, thanks - it looks v v nice!

shaloo:
Taking this up over the weekend...thx!

rgommers:
Thinking about who to list. Some thoughts:
- I like the Pandas ordering: Contributors, then Maintainers, then other teams.
- We have a Steering Council which Pandas doesn't have; I think that is less important Maintainers, so let's just list those towards the end under Governance.
- I'd be inclined to do it like Pandas and put active Maintainers in the gallery, and use Emeritus Maintainers for people no longer active.
[...]

Issue #89, opened 2019-11-24
Idea #2 for Team page


InessaPawson:
Add a snapshot of Google Maps indicating where NumPy contributors are from.

InessaPawson:
@joelachance I think the label "design" is more fitting for this issue. I'll go ahead and change it.

InessaPawson:
Looking at this issue reminds me how far I’ve come in my understanding of open source and the NumPy community. What once felt like an exciting idea now sounds like a lot of effort without lasting value.:)
The NumPy contributor community is global and fluid by nature. Keeping a map accurate would be a perpetual challenge raising contributor privacy concerns. 
Let’s close this issue to stay focused on the work that matters.

PR #205, opened 2020-03-26
Add a NumPy Team

shaloo:
NumPy Team gallery added, script to generate gallery added.
Fixes gh-45

#### Deployed URL

http://numpy-205.surge.sh/


scripts/gallery/numpy/static/css/styles.css
```diff
@@ -0,0 +1,354 @@
+body {
```
rgommers: This file is a copy, that's not desired. Can you instead merge your changes into `static/css/styles.css` or add a new CSS file here with only the rules needed for this Team page and then use both CSS files in the template?

rgommers: The same applies to the NumPy and NumFOCUS logos, we already have them in the repo.

rgommers: And `avatar.png` seems unused so can be removed.

scripts/gallery/numpy/team.md
```diff
@@ -0,0 +1,141 @@
+## NumPy Team Gallery
+
```
rgommers: This file has a mix of tabs and spaces that would be good to fix (all spaces preferred).

More importantly, I think the raw html should be a partial; `layouts/partials/team.html` would be the place to put it. You only need one for a single gallery; the same partial can then be reused under `Maintainer`, `Web Team`, etc.

scripts/gallery/team.py
```diff
@@ -0,0 +1,261 @@
...
+        the GitHub user information for them.
+        """
+
+		stored_ctx_fname = context["stored_context"]
+		if ( stored_ctx_fname != "" ):
+			with open(stored_ctx_fname) as f:
+				storedctx = yaml.safe_load(f)
+			return storedctx
+
+		context["main"]["maintainers"]["people"] = []
```
rgommers: You can take these 8 lines and put them in a function (say `def get_peopledata(kind):`), and then call that for "core", "webteam", etc. That would get rid of a lot of repeated code and be easier to read.

scripts/gallery/team.py
```diff
@@ -0,0 +1,261 @@
...
+	contrib_fname = os.path.join(source_path, "team.yml")
+	shutil.rmtree(target_path, ignore_errors=True)
+	os.makedirs(target_path, exist_ok=True)
+	sys.stderr.write("Generating context...\n")
+	context = get_context(contrib_fname, ignore_io_errors, stored_context=stored_context)
+	sys.stderr.write("Context generated successfully. \n")
+	templates_path = os.path.join(source_path,
+								  context["main"]["templates_path"])
+	jinja_env = jinja2.Environment(
+					loader=jinja2.FileSystemLoader(templates_path))
```
rgommers: some weird alignment issues in the 4 lines above

scripts/gallery/team.py
```diff
@@ -0,0 +1,261 @@
...
+				with open(os.path.join(source_path, fname)) as f:
+					content = f.read()
+				if extension == ".md":
+					body = markdown.markdown(
+									content,
+									extensions=context["main"]["markdown_extensions"]
+					)
+					content = extend_base_template(
+									body,
+									context["main"]["base_template"])
```
rgommers: more misaligned code. please only use spaces, no tabs. you should set your code editor to always convert tabs to 4 spaces for any Python code.

scripts/gallery/numpy/team.md
```diff
@@ -0,0 +1,141 @@
...
+<div class="row maintainers">
+	{% for row in main.maintainers.people | batch(9, "") %}
+		<div class="card-group maintainers">
+		{% for person in row %}
+			{% if person %}
+				<div class="card rounded">
+					<img class="card-img-top" style="border-radius: 5%;" alt="" src="{{ person.avatar_url }}"/>
+					<div class="card-body">
+						<h6 class="card-title">
+							{% if person.blog %}
```
rgommers: Design wise, I'd prefer to remove these links. Rationale:
- Some people have links and some don't, which means tiles have either black or blue text.
- The links are fairly random, sometimes it's LinkedIn, sometimes a personal blog, or yet something else. 
- Those links can easily be found some other way

scripts/gallery/numpy/static/css/styles.css
shaloo: > This file is a copy, that's not desired. Can you instead merge your changes into `static/css/styles.css` or add a new CSS file here with only the rules needed for this Team page and then use both CSS files in the template?

ok

shaloo: Will do - this was meant to be a standalone script but I guess after integration we need to remove the duplicates. Working on it.

scripts/gallery/numpy/team.md
shaloo: ok - will look into it.

scripts/gallery/team.py
shaloo: ok

scripts/gallery/numpy/team.md
shaloo: Should we use the location information in Github instead of blog?  Or just names?

rgommers: In gh-89 there was a suggestion to use a map to show everyone's location. However that's quite a bit of work and a separate issue; for now I would suggest just to use the names.

shaloo:
> Thanks @shaloo, the output looks pretty good!
> 
> Code wise there's some cleanups to do to make this easier to maintain. A lot of committed files are unneccesary. I propose to aim for the following structure:
> 
> * `scripts/gallery/numpy_team.py`
> * `scripts/gallery/numpy_team.yml`
> * `scripts/gallery/cache/numpy_team.context`
> * `content/en/team.md`
> * `layouts/partials/team_gallery.html`
> 
> As far as I can tell, that's all you need. And then the page would end up at `/team` rather than `/gallery/team`

What about _templates? Those are used by Jinja2 to spit out html (static html with no variables unlike partials).  Or are you suggesting we use partial/team.html as the template for Jinja2 translations?

scripts/gallery/team.py
```diff
@@ -0,0 +1,261 @@
...
+The rest of the items in the file will be added directly to the context.
+
+Run ./team.py --help to see how to run it.
+
+Example:
+
+	* To run the script use:
+		./team.py <Dir Path to yml file location> --ignore-io-errors
+
+	This will pull information about team members listed in yaml file into a local file called team.context. In subsequent runs, unless the team changes, you can reuse this locally stored information to render it using this script and play with layout, css etc. as shown in the next example.
```
rgommers: this line needs breaking, it's useful info but hard to read now

rgommers:
> What about _templates? Those are used by Jinja2 to spit out html (static html with no variables unlike partials). Or are you suggesting we use partial/team.html as the template for Jinja2 translations?

As I commented on https://github.com/numpy/numpy.org/issues/45#issuecomment-605680176, I missed the `jinja2` bit. I think Jinja2 usage is undesirable here, since the little it does is something Hugo can do just as well.  However, we can leave that for later in a separate issue, given that now it works.

The duplication of the CSS and images would still be good to fix though, that really does look unnecessary and it will go out of sync as soon as the other `styles.css` is changed (e.g. @joelachance  is tweaking colors right now).

shaloo:
> Thanks @shaloo, the output looks pretty good!
> 
> Code wise there's some cleanups to do to make this easier to maintain. A lot of committed files are unneccesary. I propose to aim for the following structure:
> 
> * `scripts/gallery/numpy_team.py`
> * `scripts/gallery/numpy_team.yml`
> * `scripts/gallery/cache/numpy_team.context`
> * `content/en/team.md`
> * `layouts/partials/team_gallery.html`
> 
> As far as I can tell, that's all you need. And then the page would end up at `/team` rather than `/gallery/team`

Done.  At present, master seems to have some issues - the build shows empty page so I have not been able to validate whether modifying team.md in content/en/ folder with the one I am using (that has Jinja directives) will play well with Hugo.  Once the base master builds successfully without the changes I have made in scripts/gallery area,  I will issue a PR.  The script reorg as suggested above minus team.md relocation to suggested dir has been validated - both with GitHub as well as cache/numpy_team.context data.

scripts/gallery/numpy/static/css/styles.css
shaloo: > This file is a copy, that's not desired. Can you instead merge your changes into `static/css/styles.css` or add a new CSS file here with only the rules needed for this Team page and then use both CSS files in the template?

Eliminated.

shaloo: > The same applies to the NumPy and NumFOCUS logos, we already have them in the repo.

Addressed.

scripts/gallery/team.py
shaloo: Done.

scripts/gallery/numpy/team.md
shaloo: > Design wise, I'd prefer to remove these links. Rationale:
> 
> * Some people have links and some don't, which means tiles have either black or blue text.
> * The links are fairly random, sometimes it's LinkedIn, sometimes a personal blog, or yet something else.
> * Those links can easily be found some other way

Done.

scripts/gallery/numpy/team.md
shaloo: Ralf, as discussed via email dated 03/30, I will work on the new issue to remove jinja2 dependency. https://github.com/numpy/numpy.org/issues/213  For now, the team.html is generated by numpy_team.py script in scripts/gallery and it stores generated html in Hugo's static/gallery area.  This page is linked to the about page from Hugo main page.  I have cleaned up content/en/team.md and placed in in script/gallery folder as it has jinja2 constructs.

shaloo:
> Thanks @shaloo, the output looks pretty good!
> 
> Code wise there's some cleanups to do to make this easier to maintain. A lot of committed files are unneccesary. I propose to aim for the following structure:
> 
> * `scripts/gallery/numpy_team.py`
> * `scripts/gallery/numpy_team.yml`
> * `scripts/gallery/cache/numpy_team.context`
> * `content/en/team.md`
> * `layouts/partials/team_gallery.html`
> 
> As far as I can tell, that's all you need. And then the page would end up at `/team` rather than `/gallery/team`

All these are addressed now other than last two bullets. Those will be part of https://github.com/numpy/numpy.org/issues/213 which will remove jinja2 dependency.

rgommers:
I removed `--ignore-errors`, I'm finding that that easily drops people off the gallery silently.

Would be nice if there was an `--update-cache` mode that doesn't so easily hit the rate limit and fills in the gaps.

The current cache in this PR is incomplete, I'll see if GitHub lets me update it later today.

rgommers:
The rest of changes look good, my earlier comments were addressed.

rgommers:
One other thing gh-213 will fix is that the current page doesn't have a navbar. Not a blocker though. The gallery looks good, and the current cache is updated. The GitHub rate limiting behavior is such that a single run of this script is fine, but a couple in a row is not. If this grows a lot, we may have to add an incremental mode at some point.

Merging, thanks @shaloo!

scripts/gallery/cache/numpy_team.context
```diff
@@ -0,0 +1,1059 @@
+ignore_io_errors: false
```
eric-wieser, line 1: Is this file supposed to be committed? The fact it dupicates the yaml makes me think that perhaps it shouldn't be.

rgommers: Yes, see the discussion above - it's the output of retrieving from the GitHub API, which we don't want to do all the time.

rgommers:
I deleted the `travisci.org` webhook, future PRs should only get a single trigger to `travisci.com`. Hopefully that fixes the confusing CI status.

PR #208, opened 2020-03-30
Adds Google Analytics

### LINKED ISSUE #198: Google Analytics for numpy.org
Set up an account with Google Analytics.
See the docs for integration on Hugo: https://gohugo.io/templates/internal/#google-analytics

joelachance:
Resolves #198. Attempted to test this, but I believe we'll need to test it at the numpy.org url.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -3,6 +3,7 @@ languageCode: en-us
 title: NumPy
 theme: hugo-fresh
 disableKinds: ["taxonomyTerm"]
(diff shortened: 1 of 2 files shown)


rgommers:
> Attempted to test this, but I believe we'll need to test it at the numpy.org url.

Yes I agree.

I quickly read up on synchronous vs. asynchronous tracking, looks like the latter only has advantages. So LGTM, merging. Thanks Joe!

PR #210, opened 2020-03-30
adds list.html to quiet hugo server warning.

joelachance:
Quiets this `hugo server` warning: `found no layout file for "HTML" for "section"`


joelachance:
btw, I'm not quite sure why this works. I think this has to do with Hugo's lookup order. I'm assuming it expects a `list.html` under `_default` for some reason.

rgommers:
Not sure why it works either - all lists still seem happy, so merging. Thanks @joelachance

PR #209, opened 2020-03-30
[WIP] Updates colors to only colors in the NumPy color palette

joelachance:
Updates colors to the ones from @InessaPawson in #48.
Colors are notated in styles.css and README.md

Review, and see what everyone thinks! LGTM.

PR #211, opened 2020-03-30
Removes underline on nav link hover, updates hover color

joelachance:
Not a ticket, but I like this style for nav rollover a bit better. Previously we had the 'hugo-fresh' orange with underline. This removes the underline and updates the rollover color to the lighter NumPy blue. Open to suggestions!

Screen Shot 2020-03-30 at 7 49 04 AM

### DIFF
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -79,6 +79,11 @@ em {
 
 #navbar-item {
     color: #2e3131;
+    text-decoration: none;
+}
+
+#navbar-item:hover {
+    color: #4DABCF !important;
 }
 
 #feature-card {


rgommers:
Agreed. The orange isn't one of the theme colors, and the underline isn't really needed. So I think it's this color or a slightly darker blue. I'm happy with this change as is. Will leave it open for a bit in case @InessaPawson or @MarsBarLee want to weigh in.

rgommers:
Ah, gh-209 chances text of links here to dark blue, so light blue looks good in combination with that.

InessaPawson:
Good idea, @joelachance! I prefer the Maximum Blue (#4dabcf) for the hover.

PR #218, opened 2020-04-23
Enhanced scripts to generate numpy dependency graph for various pkg

### LINKED ISSUE #217: Create dependency charts for GwPy and Pycb for GW case study
As suggested by @rossbar for #182 , we can reuse/update script/ehtim such that it can be used to generate NumPy dependency chart for any given package - in this case GwPy and Pycd.  This will help visualise the role of NumPy in GW case study and highlight it for readers.

shaloo:

Earlier, we had a script that worked for ehtim pkg dependency graph for NumPy.  The same script is now enhanced into generic one that works for any package which is dependent on NumPy.  Here is how the new script works:

Use case 1:  You already have the package deployed locally - in that case simply run gen-numpy-dep-graph.sh.  Note: Pre-requisite - you should have pipdeptree and graphviz installed on your system.

Use case 2: You do not have the package for which you need to generate NumPy dep graph.  You can use fetch-numpydeppkg.sh to download pkg sources and deploy locally along with graphviz and pipdeptree.  Next, you can run gen-numpy-dep-graph.sh followed by cleanup script which will uninstall and remove all downloaded sources from your local machine and free up space used by these packages.  

Use case 3: NumPy use - just one one script which generates updated dep graphs of all pkg that are referred to in NumPy.org case studies.  This is done by redraw-numpy-dep-charts.sh and it also serves as an example for Use case 2.

Fixes #217

#### Deployed URL


rgommers:
Note to self: the current outputs of these are in gh-182.

scripts/fetch-numpydeppkg.sh
```diff
@@ -0,0 +1,97 @@
...
+# The following function does the following:
+# 1. Fetches latest software sources for the specified package in the given  working dir
+#    (default current-dir/numpy_<pkgname>_dep)
+# 2. Assumes python3, virtualenv, python and pip3 are available and installs graphviz, pipdeptree
+# 3. Git clones specified package
+#
+# To generate the dependency graph for the specified package and highlight the
+# role of NumPy use the script gen-numpy-dep-graph.sh
+# 
+
```
rgommers, line 19: It would be good to use `set -euxo pipefail` here, so the scripts fails at the point something goes wrong. It took me a while to figure out I was missing `virtualenv` ...

scripts/redraw-numpy-dep-charts.sh
```diff
@@ -0,0 +1,74 @@
...
+#
+
+set -ue
+
+curdir=`pwd`
+outputdir=$curdir
+
+if [[ "$#" -ge "2" || (( "$#" -eq "1" ) && ( "$1" == "-h" || "$1" == "--help" )) ]]; then
+    echo ""
+    echo "Usage:  ./redraw-numpy-dep-charts.sh outputdirname <optional, default current-dir"
```
rgommers, line 13: This script actually only works when running from the root of the repo: `./scripts/redraw-numpy-dep-charts.sh`

scripts/fetch-numpydeppkg.sh
shaloo: ok, will do.

scripts/redraw-numpy-dep-charts.sh
shaloo: ok, will take care of this too.

shaloo: I'm using MacBook to test these scripts.  So far they all seem to be running fine without making any changes, yet.  Did you run them on linux?

shaloo: The new set of scripts deprecates fetch-ehtim and NumPy-ehtim-dep-graph  which were specific to generating ehtim graph only.  The new scripts are generic for specified packages albeit run in the order listed in top level script - redraw-numpy-dep-charts.sh

shaloo: Creating readme for correct usage.

scripts/fetch-numpydeppkg.sh
shaloo: done.

shaloo:
> Thanks @shaloo! These new graphs will be useful. The script does work for me, with some minor hiccups. It would be useful to add a `scripts/README.md` to make it easier to figure out what to do.

Hopefully you will have a better experience with the fixes. Pls do let me know if you have any more suggestions and inputs to make it better.

shaloo:
clicked closed by mistake, reopened.

shaloo:
I tried to squash last two updates to readme as part of this update.  Had to force push after rebase with upstream/master - hope that is ok.

scripts/redraw-numpy-dep-charts.sh
shaloo: > This script actually only works when running from the root of the repo: `./scripts/redraw-numpy-dep-charts.sh`

Addressed, thanks for highlighting this bug.

rgommers:
> Had to force push after rebase with upstream/master - hope that is ok.

No worries, force-pushing is always okay. The only rule is never to do that to a branch used by other people (e.g. `master` in this repo). On your own branches and PRs it's always fine.

rgommers:
This succeeds out of the box now doing simply `./redraw-numpy-dep-charts.sh`. Result:

image

The PyCBC figure would look cleaner with `six` removed, but let's not hold up this PR on that - it looks good too as is, we can always work on this more in the future if we want to further improve this.

rgommers:
Image file sizes here are ~100 kb, which is good.

rgommers:
Merged, thanks @shaloo!

PR #221, opened 2020-04-24
Moves shell below key features

### LINKED ISSUE #220: Revise the info architecture of the home page
Place the shell below the “Key Features” section to align better the website info architecture elements with the project objectives.

joelachance:
I've added a Title to NumPy Shell here, and right now I'm indifferent about the result. It felt like something was missing when I moved it below Key Features, so please post thoughts and suggestions.

Resolves #220


joelachance:
@InessaPawson 
1. I did notice the whitespace and had removed it. You might have cached styles. I'll double check though!
2. Without the 'Learn More', I don't think there's a CTA in that section.
3. Will do!

InessaPawson:
> 1. I did notice the whitespace and had removed it. You might have cached styles. I'll double check though!

Clearing the cache fixed it. Thank you for the tip!

> 2. Without the 'Learn More', I don't think there's a CTA in that section.

I’m not certain if the CTA is really needed there. @MarsBarLee @rgommers what do you think?

rgommers:
> It felt like something was missing when I moved it below Key Features

Yeah, it's hard to explain exactly what it is, but visually the page looks less appealing to me now. Perhaps because I'm already used to the original layout, but I'm not completely sure this is an improvement.

InessaPawson:
I'll submit a prototype with the new layout for you to make the final decision.

InessaPawson:
Here are four lo-fi prototypes for your consideration. 

**Version 1** Key features -> News -> Shell
numpy_h1
**Version 2** Shell -> News -> Key features
numpy_h2
**Version 3** News -> Shell -> Key features
numpy_h3
**Version 4** News -> Key features -> Shell numpy_h4

joelachance:
One suggestion if we go with the banner-style news section: We could do the blue background in CSS and allow that to stretch across the screen 100%, and simply export the rest of the image as an svg. On that note, if we are just rendering text, it'll be a bit more performant to write it in HTML. On the flip side, this could be a pain if this section changes a lot.

My preference would be v3 or v4.

rgommers:
I like v4 the best, and I think it will look even better with Joe's suggestion of stretching the news bar to 100% of site width. v4 has a nice symmetry to it somehow, and fixes the thing that felt a little off with the "key features" section right under the title.

> and simply export the rest of the image as an svg. On that note, if we are just rendering text, it'll be a bit more performant to write it in HTML. On the flip side, this could be a pain if this section changes a lot.

The text of that news bar will change regularly I think. The text shouldn't be inside an svg image imho, that'd be too annoying to update (unless I'm missing what you are suggesting here).

InessaPawson:
For the record, I prefer Version One slightly more. I like the logic of it. First, there is an outline of the key features of the library, then an announcement (with a link) of the latest news about the project (in this example the latest stable release), then an opportunity to try out NumPy without the installation is offered. 
Version Four would be my second choice.

InessaPawson:
> I like v4 the best, and I think it will look even better with Joe's suggestion of stretching the news bar to 100% of site width. v4 has a nice symmetry to it somehow, and fixes the thing that felt a little off with the "key features" section right under the title.
> 

Stretching out the bar would work. And if we leave out the “Learn more” button, I would not miss it.;)
> 
> The text of that news bar will change regularly I think. 
>
Correct.  @joelachance @rgommers I understand that not everyone is comfortable editing .svg. Do you think using `news` class would be a better option?

InessaPawson:
Here is Version 4 with updates.
numpy_h4_1

joelachance:
I like your v4 update, @InessaPawson!

 Why don't we put news content in the config file, and have it render in html? I think this might be much less of a chore for whomever needs to update the NumPy news.

InessaPawson:
> I like your v4 update, @InessaPawson!
> 

Great! I think we have consensus here. 

> Why don't we put news content in the config file, and have it render in html? I think this might be much less of a chore for whomever needs to update the NumPy news.

I see. It makes sense. @joelachance  Please go ahead and make changes to the code.

rossbar:
I'm a bit late, but I just wanted to chime in and say I like v4 best too, esp. after the update!

InessaPawson:
> I'm a bit late, but I just wanted to chime in and say I like v4 best too, esp. after the update!

Thank you for your feedback, @rossbar!

joelachance:
@InessaPawson @rgommers @MarsBarLee ready for another review

joelachance:
Looks like some changes from master might have gotten merged in here, hopefully those don't make this PR too confusing.
I renamed `section2` to `keyfeatures` here to keep things a bit more obvious.

rgommers:
Thanks for the update Joe. The colors and text look good. Could you make the text a link to an empty `numpy.org/news` (which can come from a `news.md`)?

There's a lot of whitespace for me and the bar is a bit on the wide side vertically:

image

rgommers:
> Looks like some changes from master might have gotten merged in here, hopefully those don't make this PR too confusing.

Oh yeah that doesn't look too good. I think it's okay for testing for now, I can help do some surgery before merging if needed - I'd prefer not to merge it with this commit history as-is.

joelachance:
agreed-- im not quite sure where that came from, or what it is. i can help remove it if everything else looks good!

rgommers:
Thanks! Yes, the rest of the changes look good to me.

InessaPawson:
> Thanks for the update Joe. The colors and text look good. Could you make the text a link to an empty `numpy.org/news` (which can come from a `news.md`)?
> 
> There's a lot of whitespace for me and the bar is a bit on the wide side vertically:
> 
@rgommers Clearing the cache should fix the whitespace issue. See Joe's tip above!;)

@joelachance The news bar is too wide indeed, let's try `min-height: 45px;`. 

Also, I suggest changing the color of the `hero-title` font and the background color of the footer to Warm Black ( #013243) to match the rest of the dark elements. (Probably, needs a separate PR.) 

These changes should make the top of the homepage look like this:
numpyorg_41

InessaPawson:
Every time I look at the homepage I can’t help but think that maybe two logos at the top one too many. @joelachance @rgommers @MarsBarLee @shaloo What do you think of removing the smaller logo just from the homepage?

rgommers:
> @rgommers Clearing the cache should fix the whitespace issue. See Joe's tip above!;)

Oops, yes that fixes it - thank you both

rgommers:
> What do you think of removing the smaller logo just from the homepage?

Dynamically hiding a logo in the header seems a little elaborate. I agree having the double logo is a little odd (but who knows, maybe it looks better with the new logo!). Maybe easier to remove the bigger one?

rgommers:
Rebased and merged in gh-228, with follow-up items noted in the PR description there. Thanks all!

InessaPawson:
> Maybe easier to remove the bigger one?

I do like the larger version of the logo next to the title NUMPY. However, if it is the only way to minimize the number of the logos on the homepage, I'm willing to sacrifice it.

rgommers:
Opened gh-229 to track the "remove a logo" todo. Thanks Inessa.

bjnath:
I've just seen the site, and it's breathtaking -- us at our best.

It's so good it has me worried. There's incredible content below the fold --
the ecosystem, the case studies -- and I'm concerned it will not get seen.

I'm wondering how we can get it higher without losing impact. Thoughts:

1) Omit the KEY FEATURES heading.

2) The features boxes can be shrunk to links -- or if we want to be fancy,
   boxes that expand on mouseover. We will not lose the supporting text; it
   will be there for those who want it.

3) `Try NumPy` is a tour-de-force -- but we have to ask the hard question of
   how much we benefit from having it on page one. The single most important
   feature of NumPy is speed, and `TryNumpy` can't show that. The section is
   less pretty and demands active engagement. On a learning page -- it's killer. 
   It's a beast. It just needs the right home.

4) I believe 1, 2, and 3 would get Ecosystem in view and with luck the Case
   Studies headline, so I'd rest easy if we stopped here. 

4a) For argument, though, I'll make a more controversial further suggestion: move
   Ecosystem over Features. Every home page plays up features, but no other
   page on the web can boast our pride of place -- the heart of a software
   ecosystem.

  After seeing that, readers wonder what's made it so damn special, and that's
  a good time to hit them with the features.

  I'd imagine a home page has two functions, the second of which is showing
  off and the first of which is getting people to another page. Once we've
  taken care of job 1, let's go all out on job 2. We can boast about our
  features, but that's not nearly as persuasive as seeing all those libraries
  we've already convinced.

joelachance:
> Rebased and merged in gh-228, with follow-up items noted in the PR description there. Thanks all!
@rgommers , thank you!

InessaPawson:
> 2. Remove the “Key Features” title and “Learn More” below.

@rgommers @joelachance Based on https://github.com/numpy/numpy.org/issues/238#issue-620392230, would you please give a second thought to the quoted suggestion? Keeping it is a waste of prime real estate indeed.

joelachance:
@InessaPawson, I'm a +1 on removing the Key Features title. I'll roll that into existing work here (assuming @rgommers agrees).
I still think a CTA somewhere in the 'key features' section is a good thing (maybe that is the tiles being links to somewhere), but I seem to remember removing the 'learn more' already (it's probably in a branch in progress, if it's not in master). I'll do that as well, pending comments from @rgommers

rgommers:
Sounds like you all agree, it seems fine with me too, the tiles will work standalone too - please go for it!

Issue #223, opened 2020-05-09
Ecosystem tab Data Science Infographic color palette update


shaloo:
Update infographics color palette to the latest primary/secondary listed here https://github.com/numpy/numpy.org/issues/33.    Ref: Infographics created for Data Science Tab under Ecosystem.

shaloo:
Closing this as changes are merged, missed tagging it during commit.

Issue #33, opened 2019-08-24
Color scheme for new site


rgommers:
From Summer, our graphic designer at Quansight, a color scheme for the new website. Top row are primary colors, second row secondary, third row tertiary. To go with the Hugo Fresh theme that we aimed for as a base. This is the same color scheme as I put on Slack earlier, copying here for permanent storage. This seems like a good start, let's see how it looks:)

![colorscheme_numpyorg_redesign](https://user-images.githubusercontent.com/98330/63642399-51f41480-c673-11e9-8d22-20315b71c83e.jpg)

shoyer:
Looks very nice!

If someone wants to refresh the colors on the NumPy logo, that would also be very welcome in my opinion. The current design looks a little faded.

rgommers:
> If someone wants to refresh the colors on the NumPy logo, that would also be very welcome in my opinion. The current design looks a little faded.

Good point. @InessaPawson's suggested the same last week - I've asked Summer to have a go at that as well.

rgommers:
There was a follow up to this where @InessaPawson posted a much more subtle color palette with several shades of blue, the ones she used in her slide template as well. I just can't find it anywhere currently.

InessaPawson:
@rgommers Here is the color scheme for the new website.

**PRIMARY COLORS**
#ffc553
#4dabcf
#4d77cf

<img width="864" src="https://user-images.githubusercontent.com/43481325/77906035-99f2ae00-72ca-11ea-9d3d-c0106767b44b.png">

**SECONDARY COLORS**
#ffffff
#013243
#eeeeee
#6c7a89

<img width="1110" src="https://user-images.githubusercontent.com/43481325/77906043-9ced9e80-72ca-11ea-9809-1e0eb396aabb.png">

sm-Fifteen:
Are these color codes relative meant to be used with an sRGB color space? There's no color profile data in the SVG logo and the SVG specification says to assume sRGB by default, so it likely is the case, but the values I'm getting looking up the names of each color are slightly different from the listed ones (though the names could be arbitrary) so I have some doubts. This isn't *critical* in most cases, but with branding it's often important to specify color-space so that the colors can be made consistent across various devices and print.

Those color codes are 3D coordinates in an RGB color space, after all, with that color space itself being defined as a polyhedron with curved faces in a different 3D space of perceived chromaticity and luminance, so the hexadecimal color codes on their own technically don't represent a specific color unless that color space is known.

[...]

PR #224, opened 2020-05-09
Fixes #gh-223 color palette update as per #gh-33 in infographics for …

shaloo:
…data science ecosystem tab

#### Deployed URL


rgommers:
Thanks @shaloo. I like the new image, colors are much cleaner and layout improved as well:

image

I did notice that both images became way larger. Is it possible to shrink them a little, >1 MB each is a bit heavy.

shaloo:
> Thanks @shaloo. I like the new image, colors are much cleaner and layout improved as well:
> 
> image
> 
> I did notice that both images became way larger. Is it possible to shrink them a little, >1 MB each is a bit heavy.

Good catch.  Fixed image size.  Also provided a way to click and view zoomed image (same image) Thanks.

rgommers:
LGTM now, merged. Thanks @shaloo!

rgommers:
One other question: what format are the originals in? It would be good to store those as well, in case we need to edit them in the future.

shaloo:
image format is png.  The same image is linked for 'click-open-in-new tab' action.  This one shows up image on full tab space of the browser window.

rgommers:
But you can't edit these right? You must have made them in some drawing program, or Powerpoint/Keynote?

shaloo:
Oh I did not get your original query.  Got it now.  These pngs are generated using powerpoint.  I have a full deck with all NumPy infographics generated so far.  It is in pptx format.  Where would you like me to check that it.  It keeps getting updated/new-items-added periodically.

rgommers:
I think adding it to https://github.com/numpy/archive would be best. Big files that get updated are fine for that repo.

shaloo:
ok, will do that right away.

PR #228, opened 2020-05-16
Moves shell below key features

rgommers:
This is a rebase of gh-221, to clean up the git history. It adds one simple commit to reduce the minimum height of the news bar (as suggested in https://github.com/numpy/numpy.org/pull/221#issuecomment-627957532).

The only things left from gh-221 are:
- Create a `news.md` page (I'll create that next), and make the news bar link to it.
- The suggestions from @InessaPawson about making the colors of the dark areas on the front page more uniform (https://github.com/numpy/numpy.org/pull/221#issuecomment-627957532)


rgommers:
Everyone seemed happy with this on gh-221, so merging. Thanks @joelachance and @InessaPawson !

PR #230, opened 2020-05-17
Add a News page

rgommers:
Changes to the news element on the front page:
- Make both title and text link to `/news`
- Tweak the text for v.1.18.0

This is what the news page looks like:

image

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -32,7 +32,8 @@ params:
   # Customizable navbar. For a dropdown, add a "sublinks" list.
   news:
     title: NumPy v1.18.0
-    content: The C-API for numpy.random defined and documented - Basic infrastructure for linking with 64-bit BLAS and LAPACK - Improved documentation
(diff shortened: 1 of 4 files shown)


rgommers:
Thanks for the review @joelachance. In it goes then.

InessaPawson:
@rgommers Happy to see the News page, but I'm not convinced that "Releases" (the release notes) belong here. They have a place on the Sphinx side, why to duplicate?

rgommers:
> @rgommers Happy to see the News page, but I'm not convinced that "Releases" (the release notes) belong here. They have a place on the Sphinx side, why to duplicate?

Well, they're kind of the most important thing we do, and the news bar on the front page also has a release focus. The Sphinx docs just have release notes, but those don't have a date. Finally, release dates are now added to https://scipy.org/, now that we have our own website we should have it there.

Issue #244, opened 2020-05-19
Rewording of gravitational waves caption


bjnath:
The original is dramatic. but the wording makes NumPy rather than gravitational waves the antecedent of "they". 

PR #245 is a replacement.

PR #245, opened 2020-05-19
Rewording for gravitational waves caption

bjnath:
Fixes gh-244

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -46,7 +46,7 @@ params:
       alttext: First image of a black hole. It is an orange circle in a black background.
       url: /case-studies/blackhole-image
     - title: Detection of Gravitational Waves
-      text: One hundred years after they were predicted by Albert Einstein, NumPy was instrumental in the first observation of gravitational waves by LIGO in 2016.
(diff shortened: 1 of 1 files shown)


rgommers:
@bjnath there's 3 commits in this PR, while you intended to submit only the last one. The reason is that you correctly created a new branch for each change, however you based the new branch off of the branch you submitted the previous PR from. What you want to do instead is base the new branch off of master, so your branches and PRs are independent of each other.

Of course if changes depend on earlier commits then that doesn't work, but that's unusual (and not the case here) and you should either keep it in a single PR or comment on the dependency.

Don't worry about it for the PR's you already submitted, I'll fix the history for those. For new ones, please try to start from master.

If this is a bit confusing please let me know!

rgommers:
LGTM, merged. Thanks @bjnath

bjnath:
> If this is a bit confusing please let me know!
Argh, on the contrary I know only too well what you mean. I realized this morning that all my PRs would have other PRs coming along for the ride. Sorry for that -- thanks for cleaning it up.  I'll revert to master for future ones.

Issue #240, opened 2020-05-19
Make shell example cut-and-pastable


bjnath:
Users will want to cut and paste the example at the left onto the shell, so it should be presented as one block, with the intervening text (e.g., "Create a 2-D array") being changed into comments.

![image](https://user-images.githubusercontent.com/6691888/82317440-40d6fa80-999c-11ea-9953-b03ca6b54cc6.png)

Playing the role of dumb user I cut and pasted just the middle section since it seemed the most interesting, and of course saw

![image](https://user-images.githubusercontent.com/6691888/82318385-c3ac8500-999d-11ea-94c3-185eed6a89e0.png)

rgommers:
That's a fair comment. I liked the look of proper text rather than code comments, but copy-pasting is useful. I can take this along in the final cleanup before launch.

PR #251, opened 2020-05-20
Make example in interactive shell copy-pastable

rgommers:
Closes gh-240

The way it looks after this change:

image

I can to tweak the wording to not make it run off the screen, and change `exponential` to `normal` because it's shorter.

### DIFF
--- a/layouts/partials/shell-lesson.html
+++ b/layouts/partials/shell-lesson.html
@@ -1,14 +1,10 @@
-
-The standard way to import NumPy:
-
 ```python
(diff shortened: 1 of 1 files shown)


bjnath:
"The standard way to import NumPy" is in regular type above the example -- do you want to incorporate it in the comments? Otherwise it seems to pertain to everything below.

rgommers:
Good point, thanks @bjnath. Fixed:

image

rgommers:
I'll merge this, it's a simple improvement.

Issue #229, opened 2020-05-16
Remove one of the two logos at the top of the front page


rgommers:
As discussion on gh-221, there are now two logos on the front page. Discussion:

>>>    What do you think of removing the smaller logo just from the homepage?

>> Dynamically hiding a logo in the header seems a little elaborate. I agree having the double logo is a little odd (but who knows, maybe it looks better with the new logo!). Maybe easier to remove the bigger one?

> I do like the larger version of the logo next to the title NUMPY. However, if it is the only way to minimize the number of the logos on the homepage, I'm willing to sacrifice it.

joelachance:
My preference would be to remove the logo in the nav bar and keep the 'exponential' logo with the NumPy title, until the user scrolls down (nav bar logo shows after user scrolls). 

There's potential that's the harder fix (although we may need to simply remove the logo config from the nav partial), and if other people prefer the other option of removing the exponential logo, then I'm game!

rgommers:
I'm happy with removing the navbar logo unconditionally.

InessaPawson:
> My preference would be to remove the logo in the nav bar and keep the 'exponential' logo with the NumPy title, until the user scrolls down (nav bar logo shows after user scrolls).
> 
+1 for the above. I thought there was a way!;) Thank you, @joelachance!

rgommers:
We merged the PR for this, closing.

PR #255, opened 2020-05-21
Bug/mobile css

### LINKED ISSUE #227: Misc Mobile rendering issues and CSS suggestions
The shell-content wiht blocks of code make the website rendering looks weird on mobile
```
.shell-content {
    min-width: 500px;
}
```

Seem to interfere wit mobile rendering issue by forcing the page to be too wide, which lead to while gap on the left:

Before/After removing the 500px min width
Screen Shot 2020-05-13 at 10 00 00
Screen Shot 2020-05-13 at 10 00 07

Can't be completely removed as the code blocks become unreadable, but I'm guessing those can be tweaked with media-screen queries

also:

```
.section3-box-content {
    margin: 30px 30px 30px 0;
}
```

Screen Shot 2020-05-13 at 10 03 28

Is not centered on mobile view either. 30px on all side is likely what you want on small screen (via media queries as well?). 

`.section2` seem to have the same issue but less visible.

There also seem to be a few screen width transition that are either off by one pixel between files or used only once:

```
// syle.css

// only transition at at 320
@media only screen and (max-width: 320px) { 

// syle.css only transition at at 400
@media only screen and (max-width: 400px) {
```

```

// section2.css only transition at at 420
@media only screen and (max-width: 425px) {
    .section2 .container {
```

Both of the following transition are off by 1px or 2px from the one in the theme used (bulma) and might lead to edge case behavior.
```
// contents.css
@media only screen and (max-width: 1088px) {
    .shortcuts-container {
        display: none;
    }
}
```

```
// section2.css
@media only screen and (max-width: 1090px) {
    .section2 .container {
        margin: 0 6vw;
    }
}
```

Chrome says there are transitions at 1087px,1088px, and 1090px maybe all those values should be collapsed ? Usually the less transitions the better has you have less screen width to test for rendering issues. These off-by-one are common even in framework as it is rarely clear in min-screen-width/max-screen-width if the end value is included or not. 

New website looks great otherwise will try to review a content and give feedback ! Good work !

joelachance:
Resolves #227, #229 
Removes Key Features title.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -56,7 +56,6 @@ params:
       url: /case-studies/cricket-analytics
 
   keyfeatures:
-    title: KEY FEATURES
     features:
     - title: Powerful N-dimensional arrays
       text:  Fast and versatile, the NumPy vectorization, indexing and broadcasting concepts are the de-facto standards of array computing today.
--- a/layouts/partials/keyfeatures.html
+++ b/layouts/partials/keyfeatures.html
@@ -1,12 +1,8 @@
 {{- $keyfeatures := .Site.Params.keyfeatures }}
-{{- $title    := index $keyfeatures "title" }}
 {{- $features := index $keyfeatures "features" }}
 
 <section class="keyfeatures">
     <div class="container">
-        <h1 class="keyfeatures-header">
-            {{ $title }}
-        </h1>
         <div class="keyfeatures-box-container">
             {{- range $features }}
             <div class="keyfeatures-box-content keyfeatures-underline">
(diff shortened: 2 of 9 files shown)


InessaPawson:
Good job, @joelachance! 
One minor fix is needed: the news bar is missing a link on this branch.

rgommers:
> One minor fix is needed: the news bar is missing a link on this branch.

This is already fixes in master and unrelated to this PR. If you're checking out this branch and want to see exactly what it looks like after merge, you have to rebase on master locally - then you'll see the news links in this case.

joelachance:
Yep, good catch, @rgommers!

InessaPawson:
> That turns a style issue into a usability issue, so better not to do that. Is it possible to hide the logo only on the front page?

Good catch, @rgommers! Great job, @joelachance! I'm very happy with this edit.

rgommers:
LGTM too now, merged. Thanks Joe!

PR #257, opened 2020-05-22
Hide the "array computing" page in the navbar

rgommers:
Leave the content for now - it needs work, but is still valuable. This content is also connected to the "Array libraries" tab on the front page.


joelachance:
LGTM, merging.

Issue #235, opened 2020-05-18
'Ecosystem' presentations and the slow reveal


bjnath:
I think that much of the the Ecosystem text works less well with readers than we intend.

 **Visualization** opens with "Data Visualization exposes patterns..." A response might be, "Wait, is NumPy a visualization program?" Another might be, "Why are they explaining visualization? I know what that is."

In this section and some others, we're setting the scene to dramatically introduce NumPy at the end.  It's great writing. But documentation is a harsh world, impatient and unappreciative of a slow reveal. 

I propose superficial changes for each Ecosystem section that will keep readers on track but preserves the excellent content.

For **Visualization**:

- Invert the paragraphs, so the first word is NumPy.
- Box the other paragraph so it's clear that it's sidebar-style content.

For **Data Science**:

- Similar treatment: start the section with the "Python has.." and "Pandas helps..." paragraphs 
- Box the Data Science overview.

For **Scientific Domains**:

> NumPy, along with SciPy and Matplotlib. are the key Python packages in multiple scientific domains. Nearly every branch of science must acquire data (from experiments or simulations), process it, and visualize the results. NumPy brings the computational power of languages like C and Fortran to Python, a language much easier to learn and use. With this power comes simplicity: a solution in NumPy is often clear and elegant. 

...followed by the picture.

For **Machine Learning**:

> NumPy forms the basis of powerful machine learning libraries like scikit-learn and SciPy. As the field expands, so does the list of libraries built atop NumPy. TensorFlow...

And again at bottom a box glossing machine learning. 

Also:

* (In citing the libraries we should follow their spelling and make `scikit-learn` lowercase and `TensorFlow` camel case.)

* (There's a text indent in the current section, next to TensorFlow, without an image.)

For **Array Libraries**:

> As hardware technology and architecture expand, new libraries are written to exploit the innovations, taking NumPy as their starting point.  CuPy, Sparse and Dask implement the NumPy API with support for  modern user cases and scalable hardware; Xarray and TensorLy add newer functionality.

Also:

* We should clarify what's meant by "*modern* user cases" or "*newer* functionality."
* Although we have links below, we should also include links in the text for consistency; all the other sections have them.

I'd be happy to put all this in a PR -- I'd need to know what to edit.

rgommers:
These all sound like good edits, thanks @bjnath.

> I'd be happy to put all this in a PR -- I'd need to know what to edit.

[...]

Issue #234, opened 2020-05-17
Reduce categories in the top menu bar


bjnath:
Some subjects in the menu bar might be hard for an outsider to tell apart. Then the menu bar fails to serve its purpose -- getting people where they want to go swiftly. 

1. Install
1. Documentation
1. Learn
1. Array computing
1. Community
1. About Us
1. Contribute 

We shouldn't force the reader to guess how we're distinguishing "Learn" from "Documentation." We should present one choice, "Documentation", and the landing page will let the reader choose one of the Procida paths: Tutorials, How-To Guides, Explanation, and Reference.

"Array computing" tells the reader hardly anything about what to expect. No one would have a reason other than curiosity to click it. I think I understand -- the subject is fundamental and overarching, so we've given it a place of its own. But it's documentation. Its readership will draw from exactly the subset of people who are reading documentation. To honor its significance we can give it prominence on the documentation page.

"Community," "About Us," and "Contribute" sound too similar.  The reader wonders what the difference is between the first two, and sites often seem to put Contribute on a Community page. I can see the value of making a special spot for contributors, and "about" is a web word people know to look for, so we can reduce this to "About us" and "Contribute."

joelachance:
I like the idea of combining some nav links, so +1 here.
Well said, @bjnath!

bjnath:
Thank you, @joelachance

InessaPawson:
Thank you for the detailed and thoughtful feedback, @bjnath! 

The Numpy community is large and very diverse. As a person in charge of the information architecture of the project website, my greatest challenge was addressing the needs of each stakeholder group while creating a pleasant user experience for all. My decisions were mostly determined by the two sources: Google Analytics (the data what people search in relation to NumPy) and the following discussion https://github.com/numpy/numpy.org/issues/43.

The analysis of the GA data showed that the most searched terms would fall under the category **“Documentation”**. That’s why, even though logically it’s close to **“Learn”**, it deserves its own button in the main navigation bar. 

Analyzing the same data set, I came to the conclusion that most of the visitors of numpy.org will be beginner users. It doesn’t come as a surprise given the fact that Data Science has become one of the most in-demand career paths for skilled professionals in recent years. That’s why most of the homepage content is geared towards beginner users of NumPy.

[...]

PR #258, opened 2020-05-22
DOC: Integrate Learn and Documentation pages

bjnath:
Adds a prominent link to the NumPy docs and clearly describes the remainder of the
page as external.

Adds a link to the contribution info at bottom and gives it a heading.

**This page is still named `learn.md`!**  Someone will need to make the plumbing changes for this to become the Documentation page.

Addresses part of issue #235.

### DIFF
--- a/content/en/learn.md
+++ b/content/en/learn.md
@@ -1,11 +1,11 @@
 ---
-title: Documentation
+title: Learn
 sidebar: false
 ---
 
-<p style="font-size: 140%;">The official NumPy documentation lives at <a href="https://numpy.org/doc">https://numpy.org/doc.</a></p>
+The official NumPy documentation lives at https://numpy.org/doc.
 
-Below is a curated collection of external resources. To contribute, see [below](#add-to-this-list).
+Below is a curated collection of external resources. To contribute, see the [end of this page](#add-to-this-list).
 ***
 
 ## Beginners
@@ -30,7 +30,7 @@ There's a ton of information about NumPy out there. If you are new, we'd strongl
 * [From Python to NumPy *by Nicolas P. Rougier*](https://www.labri.fr/perso/nrougier/from-python-to-numpy/)
 * [Elegant SciPy](https://www.amazon.com/Elegant-SciPy-Art-Scientific-Python/dp/1491922877) *by Juan Nunez-Iglesias, Stefan van der Walt and Harriet Dashnow*
 
-You may also want to check out the [Goodreads list](https://www.goodreads.com/shelf/show/python-scipy) on the subject of "Python+SciPy". Most books there are about the "SciPy ecosystem," which has NumPy at its core.
+You may also want to check out the [Goodreads list](https://www.goodreads.com/shelf/show/python-scipy) on the subject of "Python+SciPy". Most books there are about the "SciPy ecosystem", which has NumPy at its core.
 
 <i class="far fa-file-video"></i> **Videos**
 
--- a/content/en/learn.md
+++ b/content/en/learn.md
@@ -1,16 +1,16 @@
 ---
-title: Learn
+title: Documentation
 sidebar: false
 ---
 
-To learn how to use NumPy, you could begin with the resources listed under the **'Beginners'** section. Once you gain a broad view, you could try the more involved tutorials and resources listed under the **'Advanced'** section.
+<p style="font-size: 140%;">The official NumPy documentation lives at <a href="https://numpy.org/doc">https://numpy.org/doc.</a></p>
 
-This is a curated collection of NumPy related educational resources. Some are very specific to NumPy, while others offer a broader view on numerical computing. It is a continuously evolving list. In case you would like to contribute, please refer to the information at the bottom of this page.
+Below is a curated collection of external resources. To contribute, see [below](#add-to-this-list).
 ***
 
 ## Beginners
 
-There is tons of information about NumPy out there. If you are new to NumPy, we'd strongly recommend the following tutorials, books and videos to get started.
+There's a ton of information about NumPy out there. If you are new, we'd strongly recommend these:
 
 <i class="fad fa-chalkboard"></i> **Tutorials**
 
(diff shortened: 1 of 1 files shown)


rgommers:
I can't find the main discussion now, it's not gh-235.

I had another look at how other sites do this.

TensorFlow:

image

Pytorch:

image

Julia:

image

Jupyter:

image

I feel TensorFlow does it best: the split between `Learn` and `API` is clear, although it should probably be called `API docs`, and it gets you to where you want to go the fastest.

bjnath:
**Learn vs Documentation** 
It's entirely about the users. If this is common practice and people can tell the difference at a glance it's not a problem. 

I don't think the world has gotten there yet and we shouldn't be justifying our mistakes by looking at other sites' mistakes, but the call has been made, at least for now.

**Everything on the page is documentation**
As far as "clearly isn't" documentation, our own [how-to guide](https://numpy.org/devdocs/dev/howto-docs.html) calls it documentation:

>The Documentation for a software project is the set of reference, instructional, educational, informative material generated by the project developers and contributors, as well as discussions, presentations, videos and other **user-generated content**.  

Documentation typically comes from the project, but there's nothing in the word "documentation" that limits it to being project-generated.

bjnath:
Issue reference should have been gh-234, not 235; sorry for the confusion.

content/en/learn.md
```diff
@@ -1,16 +1,16 @@
 ---
-title: Learn
+title: Documentation
```
rgommers, line 2: Would like to keep this as `Learn`

content/en/learn.md
```diff
@@ -1,16 +1,16 @@
...
 sidebar: false
 ---
 
-To learn how to use NumPy, you could begin with the resources listed under the **'Beginners'** section. Once you gain a broad view, you could try the more involved tutorials and resources listed under the **'Advanced'** section.
+<p style="font-size: 140%;">The official NumPy documentation lives at <a href="https://numpy.org/doc">https://numpy.org/doc.</a></p>
```
rgommers, line 6: The larger font doesn't look so good style-wise. It's the first sentence, people will read it either way, so let's do normal font here.

rgommers: Also the `<a href=...` isn't needed; Hugo recognizes links without special markup.

content/en/learn.md
```diff
@@ -1,16 +1,16 @@
...
 
-This is a curated collection of NumPy related educational resources. Some are very specific to NumPy, while others offer a broader view on numerical computing. It is a continuously evolving list. In case you would like to contribute, please refer to the information at the bottom of this page.
+Below is a curated collection of external resources. To contribute, see [below](#add-to-this-list).
```
rgommers, line 8: Much more concise, nice. The one tweak I'd like to make is remove the double use of "below".

content/en/learn.md
```diff
@@ -30,7 +30,7 @@ There is tons of information about NumPy out there. If you are new to NumPy, we'
 * [From Python to NumPy *by Nicolas P. Rougier*](https://www.labri.fr/perso/nrougier/from-python-to-numpy/)
 * [Elegant SciPy](https://www.amazon.com/Elegant-SciPy-Art-Scientific-Python/dp/1491922877) *by Juan Nunez-Iglesias, Stefan van der Walt and Harriet Dashnow*
 
-Besides the ones listed above, you may want to check out the [Goodreads list](https://www.goodreads.com/shelf/show/python-scipy) on the subject of "Python+SciPy". Most of the books on this list are about the "SciPy ecosystem", which has NumPy at its core.
+You may also want to check out the [Goodreads list](https://www.goodreads.com/shelf/show/python-scipy) on the subject of "Python+SciPy". Most books there are about the "SciPy ecosystem," which has NumPy at its core.
```
rgommers, line 33: minor: is putting the comma within the quotes normal style? looks a little odd to me

content/en/learn.md
```diff
@@ -78,6 +78,6 @@ If you have the basic know-how of NumPy and how to use it, you can try these adv
 
 If NumPy has been significant in your research, and you would like to acknowledge the project in your academic publication, please see [this citation information](/citing-numpy).
 
-***
-
-_If you would like to contribute to this collection of NumPy Learning resources, please submit your recommendation [here via a pull request](https://github.com/numpy/numpy.org/blob/master/content/en/learn.md). Make sure that your submission includes a brief description regarding why it deserves mention on numpy.org and what audience would benefit from it the most._
+## Contribute to this list
+<a name="add-to-this-list"></a>
```
rgommers, line 82: Having the anchor to link to is nice. I'm not really wanting to encourage submissions that much though, given that I fully expect to have to reject a large majority of submissions. So a smaller section header would be better I think.

rgommers: Or maybe it doesn't matter. Resolving.

rgommers:
In it goes, thanks @bjnath

bjnath:
> minor: is putting the comma within the quotes normal style? looks a little odd to me

I was inconsistent here; the period should be inside "Python+SciPy." too.

Per [Google style guide](https://developers.google.com/style/):
image

content/en/learn.md
bjnath: > minor: is putting the comma within the quotes normal style? looks a little odd to me

I was inconsistent here; the period should be inside "Python+SciPy." too.

Per [Google style guide](https://developers.google.com/style/):
image

content/en/learn.md
bjnath: +1, missed that.

content/en/learn.md
bjnath: You would *think* the first line has prominence.  I'm concerned that because there's **so** much other material people scanning the page won't see it (think of how we look at a new page -- we scan it quickly before reading it).

Is **bold** a reasonable compromise?

rgommers:
Okay, I'll make these last suggested changes both here and on the data science tab.

bjnath:
If we're talking last changes before launch, I'd urge changing "Get started" to "Learn NumPy."

rgommers:
Ah yes, will look at that issue too.

content/en/learn.md
rgommers: that does look better. bold italics or simple italics work too. I went with bold

bjnath: Looking at the page I'm feeling better about how the Learn/Documentation issue is turning out. People who click Learn get a chance to re-route to the official docs, without the frustration of having to mentally backtrack and try another nav link.

We can use a symmetric link on the official side pointing to the external docs (or may already have another page of them), but that's less urgent.

bjnath: Just realized: That "external" page includes NumPy User Guide (last entry in Tutorials) -- should be removed.

bjnath: Also -- I'd urge that the hyperlink *not* say 'here'!  It's too easy to not notice the link and misread "here" as meaning "this page".  

Instead of `here`, the URL should be spelled out: https://numpy.org/doc/stable

It should explicitly say https://numpy.org/doc/stable so that nobody thinks "here" me

PR #259, opened 2020-05-23
Updates colors to Warm Black

### LINKED ISSUE #246: Change the color of the `hero-title` font and footer background
Change the color of the `hero-title` font and background color of the footer to Warm Black      (#013243) to match the rest of the dark elements.

joelachance:
Resolves #246

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -2,7 +2,7 @@
 {{- $logo        := index $footer "logo" }}
 {{- $quickLinks  := index $footer "quicklinks" }}
 {{- $socialMedia := index $footer "socialmedia" }}
-<footer class="footer footer-dark">
+<footer id="footer" class="footer">
   <div class="container">
     <div id="footer-columns" class="columns">
       <div class="footer-logo-column">
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -31,6 +31,10 @@ em {
     margin: 15px;
 }
 
+.hero-container {
+    width: 100vw;
+}
+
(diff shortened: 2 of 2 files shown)


joelachance:
Screen Shot 2020-05-22 at 10 44 37 PM
Screen Shot 2020-05-22 at 10 44 31 PM

rgommers:
Thanks @joelachance. I pushed a fix for the merge conflict here after merging the mobile CSS PR.

Colors look good, I do have two small suggestions:
- the link hover color in the footer turned blue now, keeping it white (same as the news bar) would look better I think
- the vertical alignment between the links section and the rightmost column is a little odd. would it be possible to align the top of `Sign up for the latest NumPy news` with `Get Help`?

joelachance:
@rgommers something happened here, do we want the footer links top-aligned? I feel like these were vertically aligned in the center previously.

I'm going to make the hover color white, but keep the underline. Does that make sense?

rgommers:
> @rgommers something happened here, do we want the footer links top-aligned? I feel like these were vertically aligned in the center previously.

Top-aligned sound good to me. I think they weren't perfectly aligned before, but the offset got more pronounced - it immediately caught my eye and didn't before.

> I'm going to make the hover color white, but keep the underline. Does that make sense?

Thanks, I like that - and that's also how the News bar works right now.

joelachance:
Ok @rgommers, we should be in good shape here.

rgommers:
Thanks @joelachance! Still a tiny bit unequal it looks like, but way better:

image

PR #260, opened 2020-05-23
DOC: Tighten ecosystem->array libraries

bjnath:
Rewrote first sentence, trimmed rest of intro. Partly
brevity, partly because words like "newer" or "modern"
might not be clear.


rgommers:
Merged, thanks @bjnath. I have some more work coming for this tab, will rebase on top of this change.

PR #242, opened 2020-05-19
DOC: Emphasize NumPy in Ecosystem openers

bjnath:
#### Brief description of what is fixed or changed

Puts NumPy in the opening sentence of each Ecosystem section,
closing issue #235.

Reduces text in the Visualization section, addressing @rgommers's concern in issue #238.

Closes gh-235.

PR #262, opened 2020-05-23
DOC: Tighten ecosystem->data science per #242

bjnath:
Proposed edit per @rgommers comment in #242:
"Still would like to make this tab a little more compact."

### DIFF
--- a/layouts/partials/data-science.html
+++ b/layouts/partials/data-science.html
@@ -8,74 +8,20 @@
         </div>
         <div>
             <p>
-                NumPy lies at the core of a rich ecosystem of data science libraries.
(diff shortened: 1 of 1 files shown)


rgommers:
After this PR:

image

Compared to before:

image

This PR is closer to the amount of content we should keep; it removes a bunch of duplication and sticks to "one figure per tab" that I think we agreed somewhere else. It does end a bit abruptly. I'd like to keep DVC and MLFlow I think, because those are quite important but less well-known, so it helps teach people. I'll try to tweak this in a follow-up PR to better capture how broad the data science space is.

rgommers:
Closed in favor of gh-272. Thanks @bjnath

Issue #261, opened 2020-05-23
Permissions for use of logos and imagery


rgommers:
There's images on some pages that we used with a caption crediting the source - those have been checked at merge and seem fine. Would be good to track other permissions to use images and logo though. Here's a list of logos, used in the "array libraries" tab on the front page (or planned to be used):

- [x] Dask - approved received from Matt Rocklin and Peter Wang
- [x] Xarray - approval by Stephan Hoyer
- [x] CuPy - approval by Crissman Loomis
- [x] Pydata/Sparse - approval by Hameer Abbasi
- [x] JAX - approved by Matthew Johnson and Paige Bailey
- [x] Xtensor - approval by Sylvain Corlay
- [x] Apache Arrow - "nominative use" permitted (see comment below) 
- [x] XND - confirmation by me (@rgommers) on behalf of Quansight
- [x] PyTorch - approved by Soumith Chintala
- [x] Tensorflow - approval received by email from Paige Bailey and Fred Alcober
- [x] MXNet  - "nominative use" permitted (see comment below) 
- [x] TensorLy - approved by Jean Kossaifi
- [x] Uarray - confirmation by me (@rgommers) on behalf of Quansight

I'm sending a bunch of messages to ask for permission - can't imagine anyone objecting, but better to do things correctly.

rgommers:
TensorFlow is the one project that has guidelines: https://www.tensorflow.org/extras/tensorflow_brand_guidelines.pdf. Relevant text:

_ATTRIBUTION  Please only use the TensorFlow name and marks when accurately referencing this software distribution, and do not use our marks in a way that suggests you are endorsed by or otherwise affiliated with Google. When referring to our marks, please include the following attribution statement: "TensorFlow, the TensorFlow logo and any related marks are trademarks of Google Inc.”_

We don't really want to include that statement, will look weird. I'll try to get an exception, will leave TensorFlow out till then.

rgommers:
Apache has a "nominative use" clause at https://www.apache.org/foundation/marks/ that allows us to use their logos without specifically asking for permission.

rgommers:
To make people's life easier, here is a current draft of this content. Final version may be different, but the main idea won't be: logo, name, short description (mostly taken from each projects' own site right now):

![image](https://user-images.githubusercontent.com/98330/82733880-568b4d80-9d17-11ea-9183-03f9cbe936c3.png)

SylvainCorlay:
[...]

PR #268, opened 2020-05-23
Update array libraries tab content

rgommers:
- removed dataframe libraries (cuDF, Xframe),
- added tensor libraries (JAX, PyTorch, TensorFlow, MXNet)
- ordered libraries more logically

image

gh-261 is the tracking issue for permissions of using these logos


rgommers:
Very minor thingy here: I cannot change "Apache" to "Arrow Apache", it doesn't fit in the column and I'm too clueless about UIkit to figure out how to give the column width a little bump.

rgommers:
Okay this is good to go, in it goes.

bjnath:
Love the wording change in Array Libraries and the new subject order.

Issue #256, opened 2020-05-21
Error when running command in Shell


joelachance:
I'm getting an error when I run `rng = np.random.default_rng()` in the shell.

rgommers:
Hmm, there's nothing wrong with that, however an old version of numpy got picked up (1.16.6) that doesn't have `default_rng`.

rgommers:
This part of `layouts/partials/javascript.html` is a bit magical:
```
          binderOptions: {
              repo: "binder-examples/requirements"
          },
```
It points to https://github.com/binder-examples/requirements which is a template that fixes numpy to `1.16.*`. We'll instead point to something we have under control; ideally a simple `requirements.txt` or inlined string rather than a full repo.

We could of course change the example here to make the error go away, but having an outdated numpy version in the shell seems undesirable.

rgommers:
The ThebeLab [docs](https://thebelab.readthedocs.io/en/latest/config_reference.html?highlight=binderOptions#configuration-reference) for this say: 

```
  // Options for requesting a notebook server from mybinder.org
  binderOptions: {
    repo: "minrk/ligo-binder",
    ref: "master",
    binderUrl: "https://mybinder.org",
    // select repository source (optional). Supports Github(default), Gitlab, and Git
    repoProvider: "github",
  },
```

I don't see an option to not provide a full repo with a `requirements.txt` (or `environment.yml` at the top level). This doesn't make much sense to me, pointing it to a simple file or, e.g., inline JSON string with requirements should be much faster. I don't see an open issue for this on https://github.com/minrk/thebelab/issues. @joelachance you've looked at this more, am I missing something?

joelachance:
@rgommers @InessaPawson 

I figured you hadn't typed the wrong command :)

I don't think you're missing anything. I just looked through the repo, and there's a lot of code around building a repo URL based on those binderOptions you hand it. I made a repo with a requirements.txt with `numpy` in it, nothing else (https://github.com/joelachance/thebelab-requirements), and that works. It was interesting watching the console when I ran it the first time, the repo is cloned, and jupyter notebook is run remotely (makes sense), so that explains the slow startup.

I did notice that the default binder-examples requirements.txt includes seaborn, matplotlib, & pandas (matplotlib we knew about), but we probably don't want those others. My repo doesn't seem to be any faster.

I'm not sure what's best here given our upcoming launch. I'd be interested in a short meeting next week to discuss next steps with the shell, given some of the issues we've had.

[...]

PR #270, opened 2020-05-23
Update numpy version in interactive shell; fix example width

rgommers:
Undo change in width of example in interactive shell. See gh-256 for details.


joelachance:
LGTM, i'll update the requirements with `matplotlib` shortly.

PR #274, opened 2020-05-24
Another edit to graphic design request for contributions

rgommers:
We now have way more design talent than we ever had, so the removed sentence was no longer representative of our situation.


rgommers:
Thanks for reviewing Inessa.

Issue #267, opened 2020-05-23
Order of tabs in ecosystem widget


rgommers:
Of all the tabs, considering both visual appeal and content, I like "data science", "scientific domains" and "array libraries" better than "visualization" and "machine learning". This is easy to switch around, especially what we choose to be first matters, because it's opened by default.

My proposed order:
1. scientific domains
2. data science
3. array libraries
4. machine learning
5. visualization

The scientific domain image is probably the one that captures best how broad the ecosystem is, hence putting that on `#1`.

Thoughts?

shaloo:
How about  array libraries before 'data science'?  That way positioning-wise  'data science' is right below 'ecosystem' and gets eyeballs first.  Also, array libraries have a big role in NumPy consumption (indirectly), that could get us hits from newer users.

rgommers:
Completely agree @shaloo, thanks. Actually I came to the same conclusion when making these changes.

PR #275, opened 2020-05-24
Switch order of tabs - "scientific domains" first

rgommers:
Closes gh-267


rgommers:
Final order is:

1.    scientific domains
2.    array libraries
3.    data science
4.   machine learning
5.    visualization

PR #279, opened 2020-05-24
Make "Documentation" link to numpy.org/doc/stable

rgommers:
The numpy.org/doc is a pretty useless page that we will get rid of in the future. The doc/stable link is new, and will keep pointing to the docs of the latest NumPy release. So let's use that.


rgommers:
Thanks, merged.

PR #272, opened 2020-05-23
Update content of Data Science tab

rgommers:
Follow up to gh-242, alternative to gh-262. The content rewrite and inclusion of more relevant libraries attempts to make this sound natural, sketch the breadth of the Python data science offerings,
and keeps some of the tools like DVC and MLFlow that beginning to intermediate data scientists really need to learn about.

It does shrink the amount of content to a more reasonable size:

image

PR #284, opened 2020-05-24
Some last minor tweaks to textual things

rgommers:
Follow up to gh-272 and gh-258


rgommers:
These were changes suggested by @bjnath, self-merging to keep things moving.

PR #286, opened 2020-05-24
DOC: Drop unneeded word from Ecosystem->Data Science

bjnath:
Minor one-word change, dropping the word "production":

>  Stable production deployments rely on

If we say "deployments" we don't need "production."


rgommers:
Makes sense, thanks

PR #287, opened 2020-05-24
Minor tweak between title and text in news bar

rgommers:
Thanks @InessaPawson for the suggestion. I agree it looks better. 

Before:

image

After:

image


rgommers:
Merging, last change before deploy.

@joelachance please double check, happy to fix this up if this wasn't quite right.

PR #289, opened 2020-05-24
Add a deploy script.

rgommers:
This follows the Hugo deploy guide for GitHub Pages, https://gohugo.io/hosting-and-deployment/hosting-on-github/#step-by-step-instructions. It worked, page is live.

This may need some tweaks to deal with cleaning up, and documentation about init'ing the submodule.

I'm also wondering if we should guard a bit better about accidental deploys, e.g. restricting permissions on `numpy.github.com` to a subset of the whole web team.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-public/
 resources/
 
 *.swp
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,5 +3,5 @@
 	url = https://github.com/StefMa/hugo-fresh.git
 [submodule "public"]
 	path = public
-	url = git@github.com:numpy/numpy.github.com.git
+	url = https://github.com/numpy/numpy.github.com.git
 	branch = master
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,7 @@
 [submodule "themes/hugo-fresh"]
 	path = themes/hugo-fresh
 	url = https://github.com/StefMa/hugo-fresh.git
+[submodule "public"]
+	path = public
+	url = git@github.com:numpy/numpy.github.com.git
+	branch = master
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+
+# If a command fails then the deploy stops
+set -e
+
+printf "\033[0;32mDeploying updates to GitHub...\033[0m\n"
+
+# Build the project.
+hugo
+
+# Go To Public folder
+cd public
+
+# Add changes to git.
+git add .
+
+# Commit changes.
+msg="rebuilding site $(date)"
+if [ -n "$*" ]; then
+	msg="$*"
+fi
+git commit -m "$msg"
+
+# Push source and build repos.
+git push origin master


stefanv:
This looks good. It won't remove files, and will keep growing the deploy repo (on our other docs deploys we keep only one commit, i.e. force a single commit with just the latest content); but that is probably not a concern, at least for now.

Deployment is automatic via Travis it looks like; in terms of guarding against accidental deploys, are you thinking of switching back to manual deploys?

rgommers:
> This looks good. It won't remove files, and will keep growing the deploy repo (on our other docs deploys we keep only one commit, i.e. force a single commit with just the latest content); but that is probably not a concern, at least for now.

I thought about it and prefer to keep history, may help with debugging, easier to go back in case of issues, and we can always clean up if size becomes a problem at some point (not expected any time soon).

> Deployment is automatic via Travis it looks like; in terms of guarding against accidental deploys, are you thinking of switching back to manual deploys?

It's still manual, TravisCI doesn't run the `deploy.sh` script I added here - it only checks that the site builds. I think we should keep it that way for a while. Also allows accumulating a set of fixes and deploying once.

stefanv:
That all sounds good. So probably just the removal of files to deal with then.

rgommers:
I'm already seeing the submodule thing is going to be annoying, I'll change it to something else. Let's not merge this. I'll keep rebasing for a little while for deploys we need until I get to a better solution for this.

stefanv:
We could fairly easily update [the NumPy docs uploader script](https://github.com/numpy/numpy/blob/master/tools/ci/push_docs_to_repo.py) to not overwrite the last commit.

mattip:
I think the simplest option would be to
- limit merges to master to content that is ready for deploy
- add the `public` directory to the repo here
- `make html` is part of the workflow, perhaps add a CI check that the content of `public` is consistent with the version/installation of hugo recommended in the instructions.
- point numpy.org to the `public` directory.

Advantages: 
- no separate repo needed
- workflow is greatly simplified.

Disadvantages: 
- inadvertent merges to master are immediately deployed, so maybe institute a "two-reviewer" policy for merging to master and lock the branch for direct pushes.
- In some sense, this repo may grow larger since each merge commits another version of the `public` content where, theoretically, the separate repo could be rewriting history to make the repo smaller. I think we can revisit this if the repo grows to take too long to clone/update at some point in the future. 

It is not clear to me how a manual process of 
- running a make step
- committing the `public` content to a separate repo
- pushing that repo

 is safer: in any case after a while the steps become force-of-habit and will be executed with the same level of caution as pushing the merge button here. In both cases, a change can be reverted by backing out the action.

rgommers:
Thanks for looking into this @mattip. I'm fine with keeping it in this repo, if so then it should be in the `gh-pages` branch though. Serving from a subfolder in `master`, if it's even possible, will make clones much slower and make it harder to rewrite history (easy to do if on a separate branch, if repo size grows too fast we'll just squash older commits).

rgommers:
I cannot remember why we have a separate repo now rather than `gh-pages`. Maybe has something to do with Cloudflare? @stefanv any thoughts?

mattip:
The [documentation for `git clone`](https://www.git-scm.com/docs/git-clone#_description) says it will clone all remote tracking branches by default, so to save cloning the `gh-pages` branch you would have to specify `--single-branch`. Another option may be to recommend cloning only part of the history via [`--shallow-since](https://www.git-scm.com/docs/git-clone#Documentation/git-clone.txt---shallow-sinceltdategt) or one of the other strategies to minimize history, that could be part of the tag line at the top of the repo description.

mattip:
The comment about rewriting history makes sense, even though this repo is only 20MB right now it could grow as time goes on. So I would be in favor of having a `gh-pages` branch. I think [in order to use github pages](https://docs.github.com/en/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site) (see step 4) you can only specify `/` or `/docs` to serve from, so the build should point hugo to create the pages in `/docs`

rgommers:
> see step 4) you can only specify `/` or `/docs` to serve from, so the build should point hugo to create the pages in `/docs`

Why not `/`? That's the regular way of using `gh-pages`. There's nothing else that should be in that branch, only the generated pages.

mattip:
Ahh, I see you are talking about [this workflow](https://gohugo.io/hosting-and-deployment/hosting-on-github/#deployment-of-project-pages-from-your-gh-pages-branch). Should I do a PR to add that deploy script to the Makefile?

rgommers:
That would be great, thanks Matti.

stefanv:
Same repo is fine; here is an example of that:

https://github.com/skyportal/skyportal/blob/master/.github/workflows/build-and-deploy-docs.yaml#L63

mattip:
@stefanv that seems to be running on repo skyportal/skyportal and publishing to skyportal/docs

I spent some time on the idea of having public be a worktree branch as in the [recommended workflow](https://gohugo.io/hosting-and-deployment/hosting-on-github/#deployment-of-project-pages-from-your-gh-pages-branch) above. It does these steps:
- deletes the public directory
- creates a gh-pages branch as a worktree in the public directory
- __rebuilds the documentation__
- commits the documentation
- now there is a worktree branch gh-pages with just the built documentation.
- pushes that branch to the original repo

Are we comfortable with building the documentation once, verifying the build, then building it again (hopefully with the same build arguments) and deploying the second build? I think a workflow like this makes more sense, since we are working with a Makefile anyway
- `make clean` will delete the public directory and delete the worktree branch in public
- `make html` will create the worktree branch in public if it does not exist and build the documentation
- `make deploy` will commit and push the worktree branch`

rgommers:
Thanks @mattip, those steps sound reasonable.

> Are we comfortable with building the documentation once, verifying the build, then building it again (hopefully with the same build arguments) and deploying the second build?

Assuming you mean the Netlify preview with the first build, yes. There are no build arguments, the whole command is simply `hugo`. I wouldn't bother with a Makefile. Just use `rm -rf public/` to clean, and `hugo` to rebuild.

rgommers:
Superceded by gh-353, so closing.

PR #291, opened 2020-05-24
Update the meta description of the site

rgommers:
This shows up at least in Slack when you simple use the link to the
site, and it didn't look pretty. Maybe this needs another change,
but for now this will do.


rgommers:
Will redeploy the site, so merging. @InessaPawson will further improve the meta fields later this week.

PR #298, opened 2020-05-24
Revised 'Install' page

bjnath:
This is a featured page in the new website, the target of Get Started. The
current version of this page is weak and the PR gets a better version out there swiftly. 
I expect a knowledgeable reviewer can improve it further.

### DIFF
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -3,7 +3,7 @@ title: Installing NumPy
 sidebar: false
 ---
 
-The only prerequisite for installing NumPy is Python, but -- particularly if you're new to the scene -- it's likely that your bigger goal is to use one of the many Python packages that depend on NumPy.  If so, see the Anaconda distribution, below.
+The only prerequisite for installing NumPy is Python, but -- particularly if you're new to the scene -- it's likely that your bigger goal is to use one of the many Python packages that depend on NumPy.  If so, see the [Anaconda](https://www.anaconda.com/products/individual) distribution, below.
 
 Package installation is Python's dark underside [\[1\]](#just-installing-stuff). If Python is new to you, we again recommend Anaconda.
 
@@ -11,7 +11,7 @@ Package installation is Python's dark underside [\[1\]](#just-installing-stuff).
            alt="Python Environment XKCD image"
            link="https://xkcd.com/1987/"
            width="400"
-           attr="_XKCD illustration - Python environment degradation_">}}
+           attr="_Image credit: xkcd_">}}
 
 
 If your results must be reproducible, the time to think about it is now; see [Reproducible Installs](#reproducible-installs).
@@ -19,10 +19,10 @@ If your results must be reproducible, the time to think about it is now; see [Re
 
 ## Anaconda
 
-Installing Anaconda gives you all the biggest Python scientific/numeric
+Installing [Anaconda](https://www.anaconda.com/products/individual) gives you all the biggest Python scientific/numeric
 players in one shot, including Python itself, which can coexist with any
 Python you already have. Of course, those packages need disk space -- about 3
-GB. If space is an issue, Miniconda is the batteries-not-included version,
+GB. If space is an issue, [Miniconda](https://docs.conda.io/en/latest/miniconda.html) is the batteries-not-included version,
 
 Packages besides NumPy that you may want to use right away include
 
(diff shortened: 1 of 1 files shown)


stefanv:
This page was much too long, so I appreciate that this PR shortens it.

I think the whole pip vs conda discussion is still too long and unnecessary; many people use pip, many use conda.  This page makes it sound like pip is a second-rate mechanism for installation, while it is the primary alternative (not the system package manager).

Why don't we provide commands for quickly creating environments using both conda and pip?  Users can copy and paste these commands, have clean environments, and be up and running in no time.

Thanks for working on cleaning this up @bjnath.

bjnath:
I meant to leave a copy of the revised page:

bjnath:
@stefanv, I appreciate the feedback!

It is undeniably an opinionated page, because its job is not to provide balanced coverage but to get users writing code.  

Installing and working in Python is different from what people are used to.  Our new users are lucky to have a pushbutton alternative to finding the packages, learning what a venv is, and getting handed a tool like pip that's likely to be misused and give grief.

One thing I particularly like about Anaconda is that you get a virtual environment whether you know what it is or not. You grow into it and eventually get the point. 

A cut-and-paste of the Python commands might initially land you in the same place, but then you're on your own remembering to use the virtual environment. Anaconda leaves no way to get it wrong.

Experienced Python users can guess the entirety of 
```
pip install numpy
```
and need nothing from the page. I'm at the other end of the curve; I found the contrast of conda and pip to be good reading.

rgommers:
Quick thoughts, I may have time to review in more detail tonight:

- This page really should start with "how to install NumPy", not "how to install Python" - it's NumPy's install page after all.
- Remember we have the install widget coming, it will replace the `conda/pip install numpy` lines that are currently at the top of the page.
- The XKCD comic should not be near the top - it's an exaggeration and not suitable as a main message.
- Yes this is opinionated, on purpose. If people don't know what they need, they are probably beginners and our advice is (and always was, see https://www.scipy.org/install.html): "install Anaconda".
- Pip definitely isn't second-rate (and advantages of it over conda over it are also listed), but it's also not a complete solution by itself, and it's not possible to give self-contained instructions for it starting from scratch.
- Good idea to add commands for creating virtual environments.

bjnath:
I think the PR supports most of these points -- or if it's not supporting them it should be.

xkcd was to hammer home that installing Python libraries requires a different mindset and can go very badly otherwise. That it's an exaggeration seems clear. If it's too ominous we can lose it.

I look forward to your comments. Honestly, I just look forward to that page being rehabilitated, by somebody. It slipped the net.

stefanv:
I think this is a significant improvement, so would recommend moving ahead.

I looked at the installation widget today, and I can probably get it done in a day or two once we have decided on exactly how it should look.  I'm not sure what we'd include, given that we have environment setup, and then also NumPy installation.  NumPy installation doesn't really need a widget, since it's:

```
pip install numpy
```

Or

```
conda install numpy
```

Other factors like conda-forge that you want to include, @rgommers?

rgommers:
> I think this is a significant improvement, so would recommend moving ahead.

There are a number of key points the currently deployed version makes (e.g. that Anaconda is for beginners, Miniconda for advanced users) that are thrown away by this PR. And there are some new statements (e.g. installing via your OS package manager is a poor choice) that are not really right.

So I don't want to merge this PR in its current form, and I won't have time to review in detail (several iterations probably) right now. The current page may need editing for conciseness and language, but what's on there now is valuable info and there's no major correctness issues that warrant being in such a hurry here.

I would suggest we start by creating and reviewing an outline (e.g. bullet points of sections and key points per section) of how this should look.

rgommers:
Re widget: I agree, could make it a few lines. E.g. including creation of virtual env via a command, and a link to more OS-specific details.

stefanv:
I think the length current page is likely to confuse and distract.  One option would be to remove the Python installation section, edit it, and return it later?

That said, I will leave it here for now and let you make the call @rgommers.

rgommers:
> I think the length current page is likely to confuse and distract. One option would be to remove the Python installation section, edit it, and return it later?

well we just moved it here because everyone else thought the separate page it was on was too hidden, and it was valuable. what you propose is how I had it till 2 days ago. I do think unhiding it was a good idea.

stefanv:
Where is 2-day older Ralf when I need him ;)

rgommers:
I was 2 days younger then:)

bjnath:
> Anaconda is for beginners, Miniconda for advanced users)

I threw it away for two good reasons: it's a) indecipherable and b) unhelpful.

Who is an "advanced user"? We don't say what that means. How is a user supposed to know if they're advanced or not?  

Is an advanced user somebody who already has installed part of the stack? Then they already have package management and don't need to be told what to use.  

Why are we making this distinction? If we're mostly targeting newcomers, the distinction they need to know is that Ana comes with everything but is a big install and Mini is bare-bones. Then for each one we tell them the best way to install it.

> And there are some new statements (e.g. installing via your OS package manager is a poor choice) that are not really right.

Happy to change that ("OS package manager comes with limitations you should be aware of.")

I worded it too strongly because I'm concerned about people installing  this way because it's familiar, without realizing what they're getting into.

> no major correctness issues...what's on there now is valuable info

Don't you think that blandly leading the page with `pip install numpy` as if it were equivalent to the conda choice is a trapdoor? 

> I would suggest we start by creating and reviewing an outline (e.g. bullet points of sections and key points per section) of how this should look.

Always a good idea.

rgommers:
> Who is an "advanced user"? We don't say what that means. How is a user supposed to know if they're advanced or not?

that's good feedback, and is fixable. depends on (a) how broad a set of packages they need, and (b) whether they're comfortable or willing to learn about conda environments from the start

> Why are we making this distinction?

It's not just size - Anaconda puts so much in the base env that the solver is literally unable to install (it will hang on large solves for 1+ hr) some extra packages that beginners will never need but advanced users likely will.

> Don't you think that blandly leading the page with pip install numpy as if it were equivalent to the conda choice is a trapdoor?

No I think it's fine. There's no real trap, users that already have Python will know if they need it or not.

rgommers:
This PR and comments on it are still a mix of good ideas, textual improvements, changes that don't look desirable, and unaddressed comments. I will cherry-pick the best parts into a new PR, and add to that the idea about providing commands for creating environments.

rgommers:
Continued in gh-373, thanks @bjnath and @stefanv

Issue #295, opened 2020-05-24
Ecosystem listing nits


rkern:
The packages for "Bayesian Inference" and "Bio Informatics" are switched.

FWIW, "Bio Informatics" should just be "Bioinformatics". I don't think anyone spells it as two words. I can understand if you want to break it up into two lines, but consider a hyphen. 

It's not clear to me why "NumPy" is under "Geographic Processing". The linked article seems out of place with all of the other links to actual packages.

PR #304, opened 2020-05-25
Improve "scientific domains" ecosystem figure.

rgommers:
Fix reported bugs, add new entries for libraries we should have included

Closes gh-295


rgommers:
This is being reported a lot, so merging and deploying.

PR #305, opened 2020-05-26
Switch sd9.png and sd10.png

InessaPawson:
Switching the images sd9.png and sd10.png to correctly illustrate the presented scientific domains.


rgommers:
Merged, thanks @InessaPawson

PR #310, opened 2020-05-27
Bug/clickable case studies

joelachance:
Fixes a small bug with the entire box being clickable for case studies.
I've also renamed `section3` to `casestudies`, since section3 probably doesn't mean much to anyone.


rgommers:
LGTM, thanks @joelachance

PR #316, opened 2020-06-03
ecosystem: add binder to Interactive Computing?

mdeff:
Cloud computing is an important part of interactive computing. Binder seems like a good representative that goes well with jupyter. CC the maintainers @betatim @choldgraf @minrk @yuvipanda.

### DIFF
--- a/layouts/partials/scientific-domains.html
+++ b/layouts/partials/scientific-domains.html
@@ -106,7 +106,7 @@
             <td class="center-text"></td>
             <td class="center-text"></td>
             <td class="center-text"><a href="https://python-visualization.github.io/folium">Folium</a></td>
-            <td class="center-text"></td>
+            <td class="center-text"><a href="https://mybinder.org">Binder</a></td>
         </tr>
         <tr>
             <td class="lastrow-center-text"></td>


rgommers:
Thanks @mdeff, good to have Binder in the ecosystem overview.

choldgraf:
Thanks! That reminds me - you guys are using Thebelab to provide kernels for the front page (which uses mybinder.org). It would be great if "mybinder.org" appeared somewhere on the page so that folks know who is running the infrastructure. For example, the spacy docs say "Launching container on mybinder.org..." when the button is clicked

rgommers:
Thanks @choldgraf, good point - adding the message in gh-323.

@joelachance those spaCy docs have a nice example of an interactive shell, when you're looking at moving to the separate page maybe there's a couple of ideas you can steal from there.

choldgraf:
@joelachance on that note, it would be great if we can brainstorm some ways to document these patterns in the `thebelab` docs so that others can riff off of them

joelachance:
@choldgraf agreed! We ended up with some 'hacky' solutions for NumPy, which I'd like to get away from.

@rgommers, I like that shell as well! We can work towards something closer to that, I think that looks nice.

PR #318, opened 2020-06-03
ecosystem: remove 3D viz from domains?

mdeff:
It's redundant to have 3D visualizations in scientific domains while there is a visualization tab (with a link to the comprehensive <https://pyviz.org>).


mdeff:
Also for consistency, as we don't have Machine Learning and Data Science in domains.

rgommers:
Fair enough, I agree that's better to remove. I had hoped that this would also make the spacing look better, because there's now seven entries in each row. It makes it look a bit worse though, on the left side:

image

I think that would be good to fix before merging this.

I don't really know right now what changes are needed for that; if you do that would be great, otherwise I can look into it later.

mdeff:
Well spotted. I think aligning groups vertically would be best, and the simplest fix would be to merge both tables in a single one. I can do it if you think that's the way to go.

mdeff:
Can we merge this? I'd proceed with a PR for #319, #320, #321.

rgommers:
Sorry for the horrible delay in merging this @mdeff. I kind of ran out of time for working on the website. Christmas holiday allowed me to finally catch up with 2020 events.

mdeff:
No problem @rgommers, you're doing a fantastic job.

Issue #315, opened 2020-06-01
Case Study: DeepLabCut - NumPy use for deep learning based  pose estimation 


shaloo:
Enhancement:  Use of NumPy for  deep learning based pose estimation  in scientific studies.

PR #324, opened 2020-06-15
Addresses #315: DeepLabCut Case Study in collaboration with Alexander…

shaloo:
… and Mackenzie Mathis

#### Deployed URL


shaloo:
Hi @rgommers, please take a look and let me know if you'd require any edits/changes. Thanks.

shaloo:
@joelachance :  Hi,  need some design guidance/ inputs regarding case-study tiles.  By default, when I add a new case study in config.yaml, it uses 3+1 time layout in two rows.  Is there a better way to add all four in a single row or may be 2 in each row?

shaloo:
@rgommers:  There are two animated gif in this case study - first one is the mice hand motion tracker (the very first image) and the other one is race horse tracking using DeepLabCut technology.  Both of them are over 1 MB and can't reduce them further.  Is it ok to use animated gifs or is it a performance overhead for our site?  In case of the latter, is there a better way to incorporate these gifs other than creating simple images for both?

rgommers:
> By default, when I add a new case study in config.yaml, it uses 3+1 tile layout in two rows. Is there a better way to add all four in a single row or may be 2 in each row?

I've also tried to tweak this, I think it's hard to get this to look optimal for every screen width. I get 2+1 by default on master with the browser window width I use, and 2+2 with this PR:

image

rgommers:
> There are two animated gif in this case study - first one is the mice hand motion tracker (the very first image) and the other one is race horse tracking using DeepLabCut technology. Both of them are over 1 MB. I tried but can't reduce them further. Is it ok to use animated gifs or is it a performance overhead for our site? 

They look really cool, so I vote to keep them! They're only loaded once one clicks on the case study tile, which will be a small fraction of all visitors. And then an extra 1.8 + 2.9 MB isn't a showstopper I think.

> In case of the latter, is there a better way to incorporate these gifs other than creating simple images for both?

The way you have done it looks good to me.

config.yaml
```diff
@@ -56,6 +56,11 @@ params:
       img: images/content_images/case_studies/sports.jpg
       alttext: Cricket ball on green field.
       url: /case-studies/cricket-analytics
+    - title: Posture Estimation using DNN
```
rgommers, line 59: I tried to tweak this title, but there's so little space......

Two minor things: I can't find a use of "posture" on the DeepLabCut, only "pose". And "DNN" I'm not a huge fan of, "deep neural networks" isn't a term used by deep learning practitioners anymore I think. Just "neural networks" is enough (I searched the TensorFlow and PyTorch websites to confirm).

`Pose estimation using neural networks` doesn't fit. `Pose estimation using deep learning` does though - what do you think about that title?

content/en/case-studies/deeplabcut-dnn.md
```diff
@@ -0,0 +1,114 @@
+---
+title: "Case Study: DeepLabCut 3D Pose Estimation"
+sidebar: false
+---
+
+{{< figure src="/images/content_images/cs/mice-hand.gif" class="fig-center" caption="**Analyzing mice hand-movement using DeepLapCut**" alt="micehandanim" attr="*(Source: www.deeplabcut.org )*" attrlink="http://www.mousemotorlab.org/deeplabcut">}}
+
+<blockquote cite="https://news.harvard.edu/gazette/story/newsplus/harvard-researchers-awarded-czi-open-source-award/">
+    <p>Open Source Software is  accelerating Biomedicine. DeepLabCut enables automated video analysis of animal behavior using Deep Learning.</p>
+    <footer align="right">—Alexander Mathis, <cite>PostDoctoral Fellow, Harvard Chan Zuckerberg Initiative (CZI)</cite></footer>
```
rgommers, line 10: I think there's no such thing as `Harvard Chan Zuckerberg Initiative`, and this should just be `Harvard`. CZI just funded one grant, but that doesn't make it an affiliation.

rgommers:
And it's up!

PR #332, opened 2020-07-11
Add news entries on 1.19.0, the new logo and the survey

rgommers:
Add three news entries, put survey on the front page (it's still running for a week):

image


rgommers:
Only one week left for the survey to run, and I'm working on deploying a new version of the site in a minute, so let's get this in.

PR #333, opened 2020-07-11
Survey end date will be dropped, so remove a mention of that

rgommers:
(after talking to @InessaPawson)

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -32,8 +32,8 @@ params:
     image: logos/numpy.svg
   # Customizable navbar. For a dropdown, add a "sublinks" list.
   news:
-    title: 1st NumPy survey is live!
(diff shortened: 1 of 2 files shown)


InessaPawson:
@rgommers I suggest the following wording in the news bar:

`title:` NumPy Community Survey is LIVE! 
`content: `Help us make NumPy better. 

And let’s link to the survey directly from the homepage. Here is a link: https://umdsurvey.umd.edu/jfe/form/SV_8bJrXjbhXf7saAl

rgommers:
Updated with the text, and also added your tweak of the logo news item @InessaPawson.

I'd prefer not to mess with the news bar linking, both because it's hard to do and because it's the only link to the whole news page.

PR #336, opened 2020-07-11
Remove Dutch translation

rgommers:
That was only ever for testing, Dutch people all are fine with English


rgommers:
and it is gone

PR #337, opened 2020-07-22
Add correct link to FEniCS in scientific-domains.html

IvanYashchuk:
The Mathematical Analysis column in scientific-domains.html had an incorrect link to FEniCS.
I have fixed that.


IvanYashchuk:
I think the intention was to link to FEniCS as it's a popular library for solving differential equations. tIGAr is an extension built on top of FEniCS.

PR #338, opened 2020-07-23
DOC: Add color swatches to NumPy palette in README

bjnath:
As a convenience, added image giving visual samples of the palette colors.


rgommers:
Thanks @bjnath. For who wants to see what this looks like:

image

I think adding an image like that is potentially useful. It should then be added as a png in the repo though I think, rather than relying on a GitHub link that may not be stable.

Also, can you just add the image rather than replace the copy-paste-able and readable-in-text-editor text?

PR #343, opened 2020-07-26
Homepage - remove the excess whitespace

InessaPawson:
@joelachance The white space in the “Key Features” section should be minimized. I’d like to hear your feedback on the proposed solution.

### DIFF
--- a/static/css/keyfeatures.css
+++ b/static/css/keyfeatures.css
@@ -1,5 +1,5 @@
 .keyfeatures {
-    margin: 30px 0;
+    margin: 10px 0;
 }
 
 .keyfeatures .container {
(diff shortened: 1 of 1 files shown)


joelachance:
@InessaPawson I personally think this is a bit too squished. I try to stick to 15px or 30px for margins as those tend to look relaxed (not too spaced out, not too tight), but of course this is all personal opinion :)

What do you think about 15px?

InessaPawson:
> What do you think about 15px?

@joelachance LGTM. I'll edit the PR and submit it.

rgommers:
15px:

image


30 px:

image

I think 15px is still too cramped. However I agree the whitespace could you a tweak, right now the spacings in different section are a little inconsistent, for example the space in the Key Features section is wider than that at the top of the Ecosystem section:

image


And then the Case Studies section title has the most whitespace:

image


A consistent application of 25px or 30px seems optimal here.

InessaPawson:
@rgommers As you know, I’m all for consistency in UI design. I’m not convinced though that the same rules should apply to every section of the numpy.org homepage. Text density and formatting, font size as well as the amount of graphic elements should be considered for striking balance in the use of white space.

rgommers:
This seems to be mostly a matter of taste, so let's merge as is and get it up. Thanks @InessaPawson, and thanks for the review @joelachance.

PR #344, opened 2020-07-28
Add PyVista to list of visualization software in the ecosystem

banesullivan:

This PR adds [PyVista](https://github.com/pyvista/pyvista) to the list of visualization software in the NumPy ecosystem. PyVista is a 3D mesh filtering and visualization library built on NumPy and the [Visualization Toolkit (VTK)](https://vtk.org). NumPy is an integral part of the foundation for PyVista and its "numpy interface" is what drives user interaction with our data types for visualization and analysis.

I also added a graphic to the "Visualization" tab. Please take a look at [PyVista's gallery](https://docs.pyvista.org/examples/index.html) and let me know if there are any other examples that might be more compelling to showcase here

Screen Shot 2020-07-27 at 10 02 05 PM

I couldn't find a way to reorganize the grid to be 3 by 3 to more evenly show all of these images now.

The image I added was created with the following code:

```py
import pyvista as pv
from pyvista import examples

# Download a volumetric dataset
# Effectively a 3D numpy array
vol = examples.download_knee_full()

cpos = [(-245.37738621091202, -98.12594361064953, 262.7952894010113),
 (74.8305, 89.2905, 100.0),
 (0.4943870362702162, -0.09406395067418555, 0.864137391594367)]

opacity = [0, 0, 0, 0.1, 0.3, 0.6, 1]

p = pv.Plotter(window_size=[144,144])
p.add_volume(vol, cmap="viridis", opacity=opacity, show_scalar_bar=False)
p.camera_position = cpos
p.show()
```

#### Deployed URL

http://numpy-344.surge.sh/

---------

cc @akaszynski and @prisae - is there a different (better?) image I should use for this? We want to keep it simple since it is 144 by 144, but also be compelling (3D volume rendering of 3D numpy arrays is pretty compelling IMO)


layouts/partials/scientific-domains.html
```diff
@@ -55,7 +55,7 @@
             <td class="center-text"><a href="https://github.com/mwaskom/seaborn">Seaborn</a></td>
             <td class="center-text"></td>
             <td class="center-text"></td>
-            <td class="center-text"></td>
+            <td class="center-text"><a href="https://github.com/pyvista/pyvista">PyVista</a></td>
```
banesullivan, line 58: I figured I'd put the git repository here as landing pages and docs are subject to change, downtime, and other variability. The github repo will always be there though

banesullivan:
@rgommers, you seem active here... is there anything I can do to move this along? Is this something you all would like to include?

rgommers:
Hi @banesullivan, thanks for the ping. Yes, this is of interest. The past weeks have been a bit of a train wreck, and this PR needs a review from me that I unfortunately didn't get around to yet. There's multiple PRs open with additions to the Ecosystem section; I still have to write up a summary of how we treat those. 

One issue when I first looked at this PR is that it added a third row to the images in the viz tab, making that look a little odd. I'd say that it would be better to get rid of one of the other images (e.g., the second one in the second row).

layouts/partials/visualization.html
```diff
@@ -50,6 +50,12 @@
                             align="middle">
                     </a>
                 </div>
+                <div>
+                    <a href="https://docs.pyvista.org/examples/index.html">
+                        <img src="images/content_images/v_pyvista.png" alt="A 3D volume rendering made in PyVista."
+                            align="middle">
```
joelachance, line 56: This `img` element is missing the closing `/`, so you'll want it like `... align="middle" />`

joelachance (changes requested):
PR looks good! Thanks for the contribution!

layouts/partials/visualization.html
banesullivan: Ah, my bad 🤦‍♂️ - will fix!

banesullivan: I just noticed none of the `img` elements on that page have a closing `/` so I added one for all

banesullivan:
I made the requested changes, now the visualization tab looks like:

Screen Shot 2020-10-14 at 10 33 18 AM

joelachance (approved):
LGTM!

rgommers (approved):
LGTM, merging. Thanks @banesullivan, and apologies for the huge delay (2020 was distracting).

layouts/partials/scientific-domains.html
rgommers: Actually I missed this line, I'd prefer to keep PyVista in the Visualization tab only, because it doesn't seem to be an image processing tool as meant in this category. 

There's a merge conflict anyway I'm resolving, I'll take out this line before merging.

banesullivan:
I'm not sure what went wrong, but the image asset included here may not have shipped to the deployed website:

Screen Shot 2020-12-28 at 9 07 44 AM

rgommers:
Hmm, very odd. Let me look into it now.

rgommers:
There was an issue in the deploy script, fixed now.

PR #351, opened 2020-08-07
Update the Docs team list

InessaPawson:
Adding @rossbar and @bjnath to the Docs team list. Fixes numpygh-349


rgommers:
Merged after adding the GitHub metadata for the cards. Thanks @InessaPawson!

bjnath:
My name isn't listed, only my GitHub id -- because my GitHub profile didn't have a name. I've added it.  Could I trouble you to rerun the script to pick up my name? 

Thanks very much.

bjnath:
Or even easier: I can PR the html for the page, and the next time the page is generated, my change will be overwritten, but it won't matter because the overwrite will be correct.

bjnath:
Opened the PR mentioned above to change the HTML, #352.

PR #353, opened 2020-08-12
deploy via gh-pages

mattip:
replaces #289 

This is based on the assumption that we still want a deploy workflow based on a local repo, not automated via CI.

The Makefile now has grown a few more options. The normal workflow would be `make clean; make html; make serve`, look at the local version or commit the changes and look at them via netlify. When a PR is merged to master,  `git checkout master; git pull; make serve`, look one more time, then `make deploy`. The pages will be put into a gh-pages branch and served via github pages from that branch.

Additionally, if you want to work on a fork, you can do all of the above with
`make TARGET=myname BASEURL=https://myname.github.io/numpy.org <command>`
which will deploy the pages to your fork, like [this](https://mattip.github.io/numpy.org/). It does take ~5 minutes for the pages to redeploy, so netlify may be a better preview mechanism. But sometimes you don't want to keep pushing little changes to this repo before squashing and opening a PR.

`make help` works and should show helpful text

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,59 @@
-.PHONY: serve html clean github
+# type `make help` to see all options 
 
-serve:
-	@hugo --i18n-warnings server -D
+# If you have naively forked this repo, say to
+# github.com://myname/numpy.org.git, you can test out the build via
+# make TARGET=myname BASEURL=https://myname.github.io/numpy.org deploy
 
-html:
-	@hugo
-	@touch public/.nojekyll
+TARGET ?= origin
+BASEURL ?= 
 
-clean:
-	@rm -rf public
+ifeq ($(TARGET), origin)
+	WORKTREETARGET =
+else
+	WORKTREETARGET = "$(TARGET)/gh-pages"
+endif 
 
-github: | clean html
-	@echo "Command to upload to git goes here"
-	@echo "See `push_dir_to_repo.py` in NumPy"
+ifdef BASEURL
+	BASEURLARG=-b $(BASEURL)
+endif 
+
+all: build
+
+.PHONY: serve html clean deploy help
+
+.SILENT: # remove this to see the commands executed
+
+serve: public ## serve the website
+	hugo --i18n-warnings server -D
+
+public: ## create a worktree branch in the public directory
+	git worktree add -B gh-pages public $(WORKTREETARGET)
+	rm -rf public/*
+
+html: public ## build the website in ./public
+	hugo $(BASEURLARG)
+	touch public/.nojekyll
+
+public/.nojekyll: html
+
+clean: ## remove the build artifacts, mainly the "public" directory
+	rm -rf public
+	git worktree prune
+	rm -rf .git/wortrees/public
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,15 +1,59 @@
...
-html:
-	@hugo
-	@touch public/.nojekyll
+TARGET ?= origin
+BASEURL ?= 
 
-clean:
-	@rm -rf public
+ifeq ($(TARGET), origin)
+	WORKTREETARGET =
```
rgommers, line 11: This does not work, it checks out `gh-pages` on the same commit as master for me. Why is the `if/else` needed at all? `$(TARGET)/gh-pages` has the correct default already.

Makefile
```diff
@@ -1,15 +1,59 @@
...
+html: public ## build the website in ./public
+	hugo $(BASEURLARG)
+	touch public/.nojekyll
+
+public/.nojekyll: html
+
+clean: ## remove the build artifacts, mainly the "public" directory
+	rm -rf public
+	git worktree prune
+	rm -rf .git/wortrees/public
```
rgommers, line 42: typo in `worktrees`

Makefile
mattip: removed.

Makefile
mattip: fixed

rgommers:
> This is based on the assumption that we still want a deploy workflow based on a local repo, not automated via CI.

Yep that's fine for now, we can change it later if needed.

In it goes, thanks Matti!

rgommers:
> Thanks Matti, this works pretty well modulo the two small issues I commented on.

Just in case, to avoid confusion when looking at this later: I actually tested this on a different Hugo site - never used it for numpy.org.

Okay I changed my mind here. It's not worth messing with Cloudflare DNS to move to `gh-pages`. Serving from https://github.com/numpy/numpy.github.com has worked fine and is actually the nicer solution (keeps this repo light), so I'll look at adapting the NumPy devdocs deploy method.

rgommers:
This is really hacky, but until the above is done, use:
```
hugo
cd public
cp -R . ../../tmp/numpy.github.com/
git commit -a
git push origin master
```

mattip:
> until the above is done

Is there something actionable that I can help with?

rgommers:
@mattip, thanks for asking - yes there is. Copying over the exact deploy method we use for the numpy devdocs to this repo, so we can deploy to the `numpy.github.com` repo.

mattip:
@rgommers if you want to deploy from the gh-pages branch, I think all you need to do is move the [CNAME file](https://github.com/numpy/numpy.github.com/blob/master/CNAME) from numpy.github.com to this repo. Then you could use [this gitub actions workflow](https://github.com/peaceiris/actions-hugo#%EF%B8%8F-create-your-workflow) to build and deploy.

rgommers:
@joelachance  looking into this, I think he identified the same GitHub Action. There's also Cloudflare settings to mess with, and it'd be nice to keep a couple of commits on the gh-pages branch so in case something goes wrong a deploy is easy to revert and debug.

PR #355, opened 2020-08-30
Update title bar after closing of the Survey

rgommers:
We didn't have v1.19.0 in the news bar yet, and it seems like the most suitable thing to put in now.


rgommers:
The survey is closed, so there's some urgency here - let me get this in before the weekend is over. Happy to make a second update if there are review comments post-merge.

InessaPawson:
No updates needed. Thank you, @rgommers!

PR #357, opened 2020-08-31
Update params.images to fix Twitter Cards (and header meta data)

joelachance:
Fixes missing images when using Twitter cards (I hope)


InessaPawson (approved):
Let's give it a try.

rgommers (approved):
Thanks Joe!

PR #359, opened 2020-09-15
make it clear this is a news item on the news page

mattip:
Add a "News:" title, use double-quotes to make hugo happy


rgommers:
Thanks @mattip. I'm happy with either the old or the new here. @InessaPawson and @joelachance, are you fine with making this change?

_old version_:

image


_version with this PR_:

image

InessaPawson:
I prefer the old version.

joelachance:
Just saw this-- I think I prefer the old version as well.

mattip:
OK, closing.

Issue #313, opened 2020-05-28
How to request inclusion in the 'Ecosystem' tab


gonzalocasas:
Following up on a [twitter conversation with numpy team](https://twitter.com/ralfgommers/status/1265765616906108934), I would like to know what's the process to request inclusion of an open source project in the 'Ecosystem' tab.

In particular, I am part of the core dev team of [COMPAS framework](https://compas-dev.github.io/), a python framework for research in architecture/engineering/digital fabrication (ie 2D/3D geometry, robotics, structural analysis, etc), but the question is also valid in general for any other project wishing to be featured there.

rgommers:
Thanks for opening this issue and suggesting COMPAS @gonzalocasas.  I'll look in more detail within the next few days.

rgommers:
Okay, not within a few days, 2020 interfered - sorry about that @gonzalocasas.

There's multiple issues about people wanting to add things to the Ecosystem tab. So here's a policy on that, which aims to help us address obvious issues, without introducing too much of a burden (we really don't want too frequent updates on categories in particular, because dealing with icons is going to be time-consuming).

_Goals for this tab are_:
1. Provide an overview that shows the breadth of scientific domains NumPy is used in. It does not aim to be comprehensive (that's impossible), just has to be a nice sampling.
2. For each included domain, highlight up to four packages that are most prominent/interesting/high-quality/large in that domain. This is a judgement call (by the NumPy web team), and easier to change. 

We now have a couple of categories with only one or two entries. We can consider replacing those with other categories. The layout has to remain appealing visually, so let's keep it at two rows of either 7 or 8 entries.

_To propose a change in categories_: open an issue to discuss first. The bar will be high here.

_To propose a change in projects listed under a category_: 
- feel free to open a PR directly.
- If the category is full (i.e. has four items), you will also need to pick the project to be removed. 
- Note that projects included should be actively maintained and in good shape, i.e. we shouldn't have doubts about recommending users to use that package.
- We will typically prefer community-driven open source projects over company-based, academic, and single-author projects.

rgommers:
Proposed new packages and categories so far:
- COMPAS Framework. That would require a fitting category to start with, which would be something like "architecture, 2D/3D geometry, robotics, structural analysis"
- gh-321 proposes to merge the signal and image processing categories
- a proposal in a https://github.com/numpy/numpy.org/issues/320#issuecomment-658979404 to move "Interactive Computing" to a separate tab (so remove the category)
[...]

Issue #369, opened 2020-11-19
Replace icons in Scientific Domains tab


MarsBarLee:
Addresses part of #342 

As of now, the icons are of varying colors, legibility, size and stroke (how thick or thin the lines are). To help unify the visual design of the site, I'm thinking of replacing the current icons with those that follow [Material Design](https://material.io/) principles.

This is a comparison of the current and some suggested icons.

---
![Icons](https://user-images.githubusercontent.com/46167686/99672687-bef68580-2a41-11eb-9759-433ca4fa980b.png)

---

![Icons Draft 2](https://user-images.githubusercontent.com/46167686/101706488-31fa8700-3a57-11eb-9158-8088b25ddcaf.png)

---

The icons are from the [Material Design website](https://material.io/resources/icons/?style=baseline) and [community created ones](https://materialdesignicons.com/), both which have Apache 2.0 licenses.

There is still some icons that don't have a currently existing icon, such as the wavelets for Signal Processing. I can make my own icons or modify existing ones to create them. I also need to adjust stroke width for some icons.

Questions: 
Are these any suggestions for the Mathematical Analysis and Scientific Computing icons? 

- Currently, they both use math symbols.

Any suggestion on the Simulation Modeling symbol?

- A brief Google image search for simulation modeling brings up a lot of flowcharts, which doesn't have a currently existing icon, so I used a tree diagram as an approximation for now.

What are your thoughts?

@InessaPawson @melissawm

MarsBarLee:
Some currently used icons feel very descriptive and fitting, such as for Geographic Processing. However, keeping the recycle/"processing arrows" would make the icon busier than the rest, creating inconsistency. I also feel like there's enough context, such as the title Geographic Processing, that simply an Earth icon would suffice.
![Geographic Processing](https://user-images.githubusercontent.com/46167686/99676924-1cd99c00-2a47-11eb-9b08-68ebd5ff1e3b.png)

rgommers:
Thanks @MarsBarLee, this looks good! I'll also Cc @shaloo, since she created the current version of this tab.

> A brief Google image search for simulation modeling brings up a lot of flowcharts, which doesn't have a currently existing icon, so I used a tree diagram as an approximation for now.

That seems very reasonable to me.

InessaPawson:
> Questions:
> Are these any suggestions for the Mathematical Analysis and Scientific Computing icons?
> 
> * Currently, they both use math symbols.

[...]

PR #374, opened 2020-12-28
Add network/graph category

jarrodmillman:
Fixes gh-320.  See https://github.com/numpy/numpy.org/issues/313#issuecomment-751466980.


rgommers:
Thanks Jarrod. Having another look around for other packages that may deserve a mention, what do you think about these:
- https://igraph.org/python/
- https://snap.stanford.edu/snappy/
- https://scikit-network.readthedocs.io/en/latest/

They all seem solid/popular/maintained, and it's not clear to me how they compare against graph-tool and PyGSP.

mdeff:
NetworkX, graph-tool, igraph, [NetworKit](https://networkit.github.io), snappy, scikit-network do [network science](https://en.wikipedia.org/wiki/Network_science): they are concerned with the analysis of networks. PyGSP, which stands for Graph Signal Processing, is more concerned with the analysis of data / features / signals supported by networks (as functions of vertices) and is closer to ([spectral](https://en.wikipedia.org/wiki/Spectral_graph_theory)) [graph theory](https://en.wikipedia.org/wiki/Graph_theory).

I have the following to say about the NetSci category:
* NetworkX is definitely the most popular
* graph-tool is well polished and maintained by its author, more mature and probably a bit more popular than NetworKit
* scikit-network is too new (first commit a year ago)
* snap and their Python SWIG interface (snappy) seems mostly used internally (by Leskovec's group at Stanford)
* igraph and NetworKit are both solid complements to NetworkX and graph-tool
* igraph is however more popular (e.g., 257,455 vs 29,153 downloads from conda-forge, or dependent projects on github and libraries.io) and mature (first commit 14 vs 7 years ago).

I see two alternative categories depending on whether or not you want to take the alternative PyGSP point-of-view:
1. Network Science: NetworkX, graph-tool, igraph, NetworKit
2. Graphs and Networks: NetworkX, graph-tool, igraph, PyGSP

I'd vote for 2, but as the author of PyGSP I'm of course biased.

rgommers:
Thanks for the thorough overview @mdeff 

> I'd vote for 2, but as the author of PyGSP I'm of course biased.

I'm happy to go with this option. "Graphs and Networks" does sound better to me.

jarrodmillman:
I updated the PR to go with option 2.

I also slightly changed the icon.  I committed the script that generates it:
- static/images/content_images/sc_dom_img/graph.py 

Feel free to change the colors, line thickess, or whatever.  Or let me know if you want me to change something.

Is there a better place to put the script?  Should I take it out of the repo?

jarrodmillman:
I just looked #369 and noticed it looks like they got rid of the colors and the grays.  I'll change the icon again.

rgommers:
Adding the script next to the icon sounds fine to me. Would it be possible to make the png size smaller though @jarrodmillman? The new icon is 10x larger than the other icons, 250 kb is a lot for a small image.

mdeff:
Why not export as SVG? Should be small given the small number of elements. I think they're moving to SVG in gh-369.

jarrodmillman:
I saved the icon as a SVG instead of a PNG.  It is now 8.0K.

I also made it 160px by 120px.  I couldn't find any specification of the preferred size, so I just randomly selected one of the existing PNGs and used its width.

rgommers (approved):
LGTM now, thanks @jarrodmillman!

PR #376, opened 2021-01-06
Add overflow: auto to .content-container to fix install page

### LINKED ISSUE #375: The "Install" page is not responsive?
Hi. First of all, thank you for the new numpy webpage!!. I love it.

When I checked the new numpy webpage from my iPhone, almost pages (front page, Learn, Community, About us, Contribute) looks responsive, but only "Install" page is not responsive. 
Also I cannot scroll the page horizontally and cannot see the right and left side of the page.

And the same thing happens in my second Android phone.
I'm not sure that this is happening in only my phones, let me report it just in case.

This is a screen shot:
IMG_9255

joelachance:
Fixes #375.

This appears to work fine with the other content pages in both mobile and desktop. This enables scrolling in the code blocks, so while you have to scroll to see the overflowing code content, the rest of the page is viewable.

### DIFF
--- a/static/css/content.css
+++ b/static/css/content.css
@@ -13,6 +13,7 @@
 .content-container {
     max-width: 850px;
     margin: 0 30px;
+    overflow: auto;
 }
 
 .content-container ul {


rgommers (approved):
Thanks Joe!

PR #377, opened 2021-01-06
Create Github Actions

joelachance:
This action uses [GitHub Actions for GitHub Pages](https://github.com/peaceiris/actions-gh-pages) to deploy to the `gh-pages` branch, using mostly the examples in that repo's readme.

This works only on the master branch. The `gh-pages.yml` config should fetch our submodules, install `0.74.3/extended`, build, and update the `gh-pages` branch.

This is supposed to work 'out of the box' with the `GITHUB_TOKEN` token-- the docs mention this is created by the Github Actions runner.

@mattip @rgommers let me know if I'm missing anything here, happy to update this pr!

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,0 +1,30 @@
+name: github pages
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  deploy:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true  # Fetch Hugo themes (true OR recursive)
+          fetch-depth: 0    # Fetch all history for .GitInfo and .Lastmod
+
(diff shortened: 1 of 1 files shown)


rgommers (approved):
LGTM, let's give this a go. Thanks @joelachance.

After we try this on the next PRs that are merged, let's verify it all looks good and then have a look at switching DNS. It would be nice to keep a few commits in the `gh-pages` branch in history for easy debugging/reverting as we discussed - we can do that later though.

rgommers:
> It would be nice to keep a few commits in the `gh-pages` branch in history for easy debugging/reverting as we discussed - we can do that later though.

Actually the action does do that already. We'll have to clean up and somehow limit the number of commits once it grows too large.

rgommers:
Follow-up in gh-379 to add CNAME. A diff on the gh-pages branch and the `numpy.github.org` repo looks good other than that.

PR #378, opened 2021-01-06
Fix Font Awesome Icons

joelachance:
Update the CDN link to Font Awesome and updates some of the icon classes so they'll render again.
Fixes the email newsletter form in the footer.
Removes text-decoration on the numfocus 'donate' button.


rgommers (approved):
LGTM, thanks @joelachance

PR #379, opened 2021-01-09
Add CNAME during deployment to gh-pages

rgommers:
Follow-up to gh-377

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -28,3 +28,4 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./public
+          cname: numpy.org


mattip:
I wonder how that will play out with the [CNAME in numpy.github.com](https://github.com/numpy/numpy.github.com/blob/master/CNAME)? Will someone "turn off" github pages serving from that repo once this goes live?

rgommers:
> I wonder how that will play out with the [CNAME in numpy.github.com](https://github.com/numpy/numpy.github.com/blob/master/CNAME)? Will someone "turn off" github pages serving from that repo once this goes live?

There's a DNS entry for `numpy.org` (well a few related ones actually, in Cloudflare), which now points to `numpy.github.com`. That needs to be switched over. There cannot really be a conflict I think.

joelachance:
LGTM! Merging.

PR #391, opened 2021-02-03
Move hugoVersion in Github Actions to correct job.

joelachance:
Let's try this, I think I had the `hugoVersion` in the wrong section of this template.


mattip:
LGTM. I cannot merge in this repo.

PR #392, opened 2021-02-03
Re-arrange Github Actions yaml

joelachance:
Remove ticks around hugoVersion, re-arranges values to match the docs

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -21,4 +21,3 @@ jobs:
           branch: master
           repo: numpy/numpy.github.com
           hugoVersion: extended_0.74.3 # Locked version, see issue #363
-          siteDir: /github/workspace/public
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -17,8 +17,8 @@ jobs:
       - name: Publish Site
         uses: chabad360/hugo-gh-pages@master
         with:
-          branch: master
-          hugoVersion: 'extended_0.74.3' # Locked version, see issue #363
-          repo: numpy/numpy.github.com
-          siteDir: /github/workspace/public
           githubToken: ${{ secrets.PERSONAL_TOKEN }}
+          branch: master
+          repo: numpy/numpy.github.com
+          hugoVersion: extended_0.74.3 # Locked version, see issue #363
+          siteDir: /github/workspace/public


joelachance:
I think we're going to need to add `PERSONAL_TOKEN` to the secrets, but I'm not sure what that's supposed to be just yet. I need to look into that.

mattip:
Did you see https://github.com/peaceiris/actions-hugo#%EF%B8%8F-create-your-workflow which does not use docker?

joelachance:
I did! We were using that action previously before the first merge in this group. I wasn't finding any config to push to a different repository, so I was going for the easiest thing. We could stick with the peaceiris action, but we may need a token anyway to push the built code into a new repo, right?

mattip:
Can we create a org-wide token that works in both repos?

joelachance:
Ok, two things, @mattip @rgommers:
1. I removed `siteDir`. I don't think we need it since it builds into the root directory (i think?)
2. I added a personal access token from my account. it only has org permissions, nothing else. There's no way to generate an org token from what I've been finding online, it's not a thing for some reason.
 
It seems it made it past the githubToken error and died on attempting to `cd` into the siteDir, so I'm hoping these changes will work.

rgommers:
> There's no way to generate an org token from what I've been finding online, it's not a thing for some reason.

Indeed. I believe the other way to do this is to create a dummy user (e.g., `numpy-deploy-bot`) and use that for such purposes. But it doesn't matter too much, happy to go with your token.

rgommers (approved):
Thanks @joelachance, let's give this a try.

joelachance:
Sounds good! I don't mind, if something weird comes up and we need to change the token, i don't see that being a big problem.
Hopefully this works 🤞

rgommers:
Still failing:
```
remote: Permission to numpy/numpy.github.com.git denied to joelachance.
```
Let me check that now.

rgommers:
Hmm, you have write access and there's no branch protection rules on that repo. So it should have worked. That's the second time, after @mattip not being able to merge PRs on this repo. Not sure what is wrong.

rgommers:
Ah, understood Matti's issue. Apparently permissions changed, this now requires full admin permission:
```
Merge pull requests on protected branches, even if there are no approving reviews
```
@mattip next time if you approve first, you should be able to merge I believe.

rgommers:
I enabled branch protection on the `numpy.github.com` repo, requiring linear history, and upgraded @joelachance's permissions from "write" to "maintainer" - I think that should fix it.

rgommers:
Same thing again, didn't fix the push permissions. I'll try a bit more.

rgommers:
Okay I changed to token to one I generated, and that fixed it.

PR #396, opened 2021-02-09
Update .content-container

### LINKED ISSUE #395: Page width issue
After updating to deployment with the GitHub Action, it looks like there's an issue with page widths - there's a scroll bar at the bottom of every page:

image

I checked if this could be due to the Hugo version used,seems unlikely, it's 0.74.3 in the Action now and I can reproduce it with v0.80.0 locally. 

Not sure what's going on there. @joelachance any idea?

joelachance:
Replace overflow: auto; with overflow: hidden. This is for content pages, as a horizontal scroll was appearing at the bottom of the page.

Resolves #395

### DIFF
--- a/static/css/content.css
+++ b/static/css/content.css
@@ -13,7 +13,7 @@
 .content-container {
     max-width: 850px;
     margin: 0 30px;
-    overflow: auto;
+    overflow: hidden;
 }
 
 .content-container ul {


rgommers (approved):
LGTM, thanks Joe!

It seems so simple when you know what you're doing:)

PR #403, opened 2021-02-17
Replace icons in Scientific Domains tab

MarsBarLee:
Resolves #gh-369

Removed Interactive Computing, Simulation Modeling and Multi-variate Analysis. Added Chemistry, Geoscience and Robotics.

Preview:

Untitled

Right now I've left the projects under the new categories Geoscience and Robotics blank since discussion about this is continuing in gh-313.


MarsBarLee:
@InessaPawson The color of the SVGs is black, hex code #00000. In the .svg files, it's stroke="#000". Does that specify the color profile (sRGB)?

InessaPawson:
@MarsBarLee  It looks much better now. I have only two suggestions:
1. The outer angles on the Math Analysis icon look cut off. Consider matching them to the icons for Robotics and Quantum Computing.
2. Consider matching the stroke width in the Graphs and Networks icon to the rest of the set.

InessaPawson:
> @InessaPawson The color of the SVGs is black, hex code #00000. In the .svg files, it's stroke="#000". Does that specify the color profile (sRGB)?

@MarsBarLee It would be better if you specify the color profile as follows:
stroke=“rgb(0,0,0)” 

Thank you for taking care of it!

MarsBarLee:
Here are the updated icons:
image

@InessaPawson 
For specifying the color space to sRGB, I was able to some SVGs from fill:"#0" to fill:"rgb(0,0,0)".

However, other ones such as 'bioinformatics.svg' don't have a fill or stroke property with any color value. It just lists the path location, such as "<path d="M4,2H6V4C6,5.44..." I think it's defaulting the color to black.

I'm looking at W3's [color property](https://www.w3.org/TR/SVG11/color.html#ColorProperty), so maybe I could add that property to explicitly define the color space? I can't find much examples online of that yet, seems like if the color isn't specified in the SVG, then the person is defining the color in CSS.

InessaPawson:
> Here are the updated icons:

@MarsBarLee Great job on the latest changes to the icon set!

> For specifying the color space to sRGB, I was able to some SVGs from fill:"#0" to fill:"rgb(0,0,0)".
> 
> However, other ones such as 'bioinformatics.svg' don't have a fill or stroke property with any color value. It just lists the path location, such as "<path d="M4,2H6V4C6,5.44..." I think it's defaulting the color to black.

Yes, under the current defaults, all SVGs are rgb(0,0,0).

> I'm looking at W3's [color property](https://www.w3.org/TR/SVG11/color.html#ColorProperty), so maybe I could add that property to explicitly define the color space? I can't find much examples online of that yet, seems like if the color isn't specified in the SVG, then the person is defining the color in CSS.

Indeed, most SVGs you'll find around the web use inline CSS, but there are some disadvantages associated with it. For example, you can’t always migrate a CSS file with an SVG, even if you know you should. For that reason alone, relying on CSS isn’t a good strategy for color management. Explicitly defining the color space and color in the SVG file is better in the long run.

InessaPawson:
@MarsBarLee My last suggestion for the icon set would be checking the width of the Signal Processing icon. It looks slightly wider than the rest of them.

MarsBarLee:
@InessaPawson I've adjusted the width of the Signal Processing icon and added the color space for the SVGs.

For defining the color space, I added _style="stroke:rgb(0,0,0)"_ to individual SVG paths. However, some icons had multiple paths, so I had to add _style="stroke:rgb(0,0,0)"_ manually multiple times. I'm saw that I could group with the [< g > tag](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/g) so that multiple paths under the group can be defined group at once. I didn't add the < g > tag because I didn't want to change the files too much yet. 

I'm working on a NumPy Docs graphics style + best practices guide so I might recommend the < g > tag or suggest people set their color space to RGB in graphic editors like Inkscape.

rgommers (approved):
LGTM, thanks a lot @MarsBarLee! And thanks for the reviews @InessaPawson!

I added new entries to resolve all the open request for adding more packages to some categories. In it goes.

Issue #398, opened 2021-02-09
Specify the color space (sRGB) in design docs


InessaPawson:
Specify the color space (RGB) and change colors to RGB in all design documentation pertaining to numpy.org and the NumPy branding assets.

PR #405, opened 2021-03-11
Specify RGB values in the docs regarding the NumPy website color palette

aerikpawson:
Specifying RGB values in the docs regarding the NumPy website color palette. Partially fixes #numpygh-398 numpy/numpy.org#398.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,19 +50,20 @@ if you don't have the Hugo extended version.
 
 ### NumPy Color Palette
 
-![#013243 Warm Black](./static/images/content_images/swatch_013243_warm_black.png) `#013243` **Warm Black**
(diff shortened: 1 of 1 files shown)


rgommers (commented):
Thanks for tackling this issue @aerikpawson!

The added RGB values all look good. It would be nice to keep the color names I think. Is there a reason why you removed them?

aerikpawson:
@rgommers  The names of the colors were removed not to mislead users since these names can change. For example, the color with the RGB value of 1/50/67, originally known as “Warm Black”,  is now called “Gunmetal”; the color with the RGB value of 238/238/238, formerly known as “Isabelline”,  can be found named as “Cultured” or “Whisper” depending on what sources you use.

rgommers:
Ah okay, fair enough - thanks for the explanation. Merging then. Thanks again @aerikpawson

aerikpawson:
Cool. Thank you, @rgommers!

PR #408, opened 2021-03-28
Update crowdin.yml files to ignore, yml format change

rgommers:
For the format change, see https://support.crowdin.com/configuration-file/#ignoring-files-and-directories

### DIFF
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -5,5 +5,8 @@ preserve_hierarchy: true
 files:
   - source: /content/en/*.md
     translation: /content/%two_letters_code%/%original_file_name%
+    ignore:
+      - /content/en/terms.md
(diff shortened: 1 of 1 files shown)


rgommers:
Actually this is still incomplete, the whole thing changed: https://support.crowdin.com/configuration-file/#configuration-file-structure

Crowdin doesn't seem to have a validator, which is annoying.

rgommers:
They just changed the whole thing to JSON, which is valid but less readable YAML. Pasting the example into http://www.yamllint.com/ will translate it back to the more human-readable form. So ignoring the docs, changing as little as possible here.

rgommers:
This didn't seem to have effect on files already in the interface for translation. I had to manually delete `terms.md`, which took effect immediately:

image

PR #410, opened 2021-03-31
Update learn.md

axil:
Added "Numpy illustrated" article.

One of the most popular numpy articles on hackernews: https://hn.algolia.com/?q=numpy


rgommers (approved):
Very nice tutorial, thanks for writing that @axil. I'd be happy to add it as a reference here. I do want to hold off on merging for a bit, since we're about to release translations into Japanese and Brazilian Portuguese that are finally at 100% - so I'll do that first.

rgommers:
Actually the translations are a bit more work than anticipated, so let's get this in now. Thanks again @axil

axil:
Thank you, @rgommers! It is an honor for me to be listed here.

PR #413, opened 2021-04-10
Integrate Japanese and Brazilian translations

rgommers:
This PR:

- [x] adds Japanese and Brazilian translations from Crowdin
- [x] enables the language dropdown in the top navbar
- [x] fixes all image link paths to work with translations
- [x] updates `config.yaml` so front page content becomes translatable
- [ ] ~extracts English content from `layouts/partials/` into `config.yaml` keys (don't want translations directly in partials)~ EDIT: postponed till next PR
- [x] refactors the `config.yaml` to be concatenated at build time from language-specific yaml files

Rather than one massive `config.yaml`, I think the optimal structure is:
```
config.yaml.in  # contains the language-independent settings
content/en/config.yaml
content/pt/config.yaml
content/ja/config.yaml
```
and then we generate `config.yaml` from the `yaml.in` template when invoking `make html`. The advantages are:
1. Keep all language-specific content in one directory, good for maintenance and enabling/disabling more languages
2. We can do the translations in Crowdin
The one downside I see is that `hugo serve` won't auto-refresh when content is changed directly in a `config.yaml` file. But that's a minor thing probably. @joelachance thoughts?

See https://gohugo.io/content-management/multilingual/ for docs on internationalization.


rgommers:
Still some more things to translate, but it's starting to look good :)

image

rgommers:
> and then we generate `config.yaml` from the `yaml.in` template when invoking `make html`.

Pretty sure the Netlify preview will work, but it requires the extra CI step to be committed already. Leaving it here for feedback on the structure. Having to type `make hugo` instead of plain `hugo` is a slight annoyance, but probably not more than that.

rgommers:
Okay, only extracting the content from the partials left to be done. I'm thinking it may be better to put the content from those in a separate yaml file (e.g. `content/en/ecosystem.yaml`) rather than make the `config.yaml` file really long).

rgommers:
Netlify is annoying; the build command config can only be changed in the Netlify UI, and that is tied to a single person's account - not me right now, so I can't change it. So I'll commit the generated `config.yaml` for the time being.

I will merge this so the translatable files show up in Crowdin. Review comments still welcome, we can always shuffle things around later. Also, extracting English snippets from layout partials is still TODO in a follow-up PR.

PR #415, opened 2021-04-17
Update config.yaml for Japanese page

AtsushiSakai:
#### Brief description of what is fixed or changed

I updated config.yaml for Japanese page.
Ref: #55

PR #418, opened 2021-04-19
Added translation of main page to Portuguese (Brazilian)

melissawm:
Some content is missing; the text in/around the interactive shell, and all content under the Ecosystem tab. ~I may be missing something but I don´t know where to find it...~ So I´ll mark this PR as a draft for now.

EDIT: it seems like the content under `layouts/partials` should be on crowdin.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -279,7 +279,7 @@ languages:
           links:
             - text: Ajuda
               link: /pt/gethelp
-            - text: Termos de uso
+            - text: Termos de uso (EN)
               link: /pt/terms
             - text: Privacidade
               link: /pt/privacy
--- a/config.yaml
+++ b/config.yaml
@@ -12,7 +12,7 @@ markup:
 # Uncomment this to disable the translations for the specified languages.
 # This may be useful to be able to merge partial translations without deploying
 # them to numpy.org immediately.
-disableLanguages: ["pt", "ja"]
+# disableLanguages: ["pt", "ja"]
 
 params:
   images:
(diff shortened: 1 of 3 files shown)


rgommers:
> EDIT: it seems like the content under layouts/partials should be on crowdin.

Yep, I know. There's various phrases scattered around in html templates that we still need to extract. We can merge the translations if `content/pt/config.yaml` is complete though (minus the `disableLanguages` change), I think that would be helpful. Are you happy with that file?

melissawm:
Ah - it seems like I did something wrong, but I don't really know. I've translated the strings in the `# Portuguese` section in `config.yaml`, and building the site I can see the complete translation. But I didn't translate `content/pt/config.yaml` (which doesn't seem to affect translations at all despite showing the same content as the former file).

rgommers:
That's because you need to run `python gen_config.py` to update the main `config.yaml`. Netlify has a very annoying permissions model, so I can't actually set it up in preview there without completely redoing that preview setup.

Let me try and fix this PR.

rgommers:
Question about the footer. "Roadmap" isn't translated, while "Terms of use" is - both are links to English pages. What do you think, which of these is best:
- _Termos de uso_
- _Terms of use_
- _Termos de uso (EN)_

image

melissawm:
Just saw your comment on #415 - I had missed this step!

Yes - roadmap is a common term and I feel like it shouldn't be translated (there is no real exact translation for it in portuguese)

`Termos de uso (EN)` sounds appropriate - I'll remove draft status for this PR, I think this is done for now. Thanks!

rgommers:
Merged, thanks Melissa!

PR #422, opened 2021-04-24
Upgrade Hugo from 0.74 to 0.82

rgommers:
This is what I use locally, seems to work quite well.


rgommers:
Ah, I cannot see what the preview job uses - need to move that over to my Netlify account.

PR #427, opened 2021-04-24
Remove generated config.toml, show translations in preview

rgommers:
This uses an env var `NUMPYORG_WITH_TRANSLATIONS` to control whether or not the site is built with translations. This works locally and in the Netlify preview, so it's no longer needed to manually toggle the setting.


rgommers:
This works, merging.

PR #446, opened 2021-06-05
Updated Installations in ReadMe.md

### LINKED ISSUE #445: Update the README.md file. (Add extra prerequisites to download MinGW)
In the Readme file, It is mentioned that you only need to install `HUGO`. 
but when the developer will move forward he/she will have to run `make hugo` and `make` command is not supported by Windows(No idea of other operating systems).

So, there should be an extra dependency to download [MinGW - Minimalist GNU for Windows](https://sourceforge.net/projects/mingw/).
or they follow [this](https://www.youtube.com/watch?v=taCJhnBXG_w) video.

Uttam-Singhh:
Resolves #445 
Updated Installations in ReadMe File, Added `python gen_config.py `

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,14 +24,14 @@ git submodule update --init
 The development web server is started with:
 
 ```bash
-make hugo
-hugo server
+make hugo 
+#or
+python gen_config.py
 ```
-
-or
-
+You can either use ```make hugo``` or ```python gen_config.py```
 ```bash
-make hugo
+hugo server 
+#or
 hugo server -D
 ```
 
--- a/README.md
+++ b/README.md
@@ -24,18 +24,25 @@ git submodule update --init
 The development web server is started with:
 
 ```bash
-make hugo 
-#or
-python gen_config.py
+make hugo
+hugo server
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,14 +24,14 @@ git submodule update --init
 The development web server is started with:
 
 ```bash
-make hugo
-hugo server
+make hugo 
+#or
+python gen_config.py
```
rgommers, line 29: This looks very confusing. Why not leave the existing instructions alone, and add below it:

_If you don't have `make` available (e.g., on Windows), use this instead_:
```
python gen_config.py
hugo
hugo server
```

rgommers (changes requested):
Thanks @Uttam-Singhh. Splitting instructions like that makes the text above it wrong, so I suggested an alternative.

README.md
Uttam-Singhh: oh yes! This is actually better I will do the changes asap!

Uttam-Singhh: Done! @rgommers

rgommers (approved):
LGTM now, thanks @Uttam-Singhh

PR #447, opened 2021-06-22
Add a page with the 2020 NumPy community survey results

InessaPawson:
Adding a page with the 2020 NumPy community survey results.


rgommers (approved):
Thanks Inessa! 

I added the pdf to this repo, and added the image of the cover to the page. It LGTM now. 

There is no entry under News, so there's no way to navigate to the `user-survey-2020` page. This is fine though until the announcement is made tomorrow morning. Let's merge this as is, you can announce it, and then we follow up with another PR which updates the News page with an entry. We also need to add an entry for NumPy 1.21.0, which was released a few hours ago:)

PR #453, opened 2021-07-23
Make an announcement about the 2021 NumPy survey

InessaPawson:
Making an announcement about the 2021 NumPy survey.


rgommers (approved):
LGTM too, thanks @InessaPawson!

Issue #370, opened 2020-11-26
Add content on performance (e.g. benchmarks, mention accelerators)


rgommers:
Related to https://github.com/numpy/numpy.org/issues/308#issuecomment-634612765 (connect content to "key features" on front page).

Adding content on benchmarks and accelerators (i.e. Cython, Numba, Pythran, Transonic) was also just suggested in https://mail.python.org/pipermail/numpy-discussion/2020-November/081248.html

melissawm:
Do you have pointers on how those benchmarks should look like? Is there a preferred set of problems to test code in? See, for example https://julialang.org/benchmarks/

rgommers:
The idea of that Julia page is about right I think, short and with only a plot, no code. The main things I'd change from that:

- fewer languages (really it's C, Fortran, Julia, R that matter)
- keep pure Python and NumPy separate
- add accelerators (Cython, Numba, Pythran, Transonic)
- change it into a couple of sections, with one plot each. There must at least be a difference between vectorize-able problems and ones where NumPy isn't such a good fit

Off the top of my head I'm not sure about an existing widely used set of benchmarks to adopt.

paugier:
Thanks @rgommers for opening this issue!

Few remarks:

- I guess it's better to keep things reasonably simple in this page so that people can have a quick overview of what can be done. I wouldn't consider too many problems.

- I think it is better to consider full problems of existing benchmark games (for example http://initialconditions.org/ or https://benchmarksgame-team.pages.debian.net/benchmarksgame/, code [here](https://salsa.debian.org/benchmarksgame-team/benchmarksgame#what-else)) and not only tiny micro-benchmarks (like in https://julialang.org/benchmarks/) to see the code in quasi-real-life situations (meaning not only few functions defined in a Jupyter notebook). It would be interesting to also mention other aspects than elapsed times, for example readability, size of the files, technical difficulties, time of coding, maintainability, etc. Optimizing is always a balance.

- One advantage of Python is that it's possible to go steps by steps from very simple implementations (sometimes not very efficient) to more complex (and more efficient) ones. It would be nice to be able to show that. The N-Body problem is a good example.

[...]

PR #458, opened 2021-07-31
Add content on performance benchmarks.

khushi-411:
This PR adds content on performance. Follows issue: https://github.com/numpy/numpy.org/issues/370

Please have a look. Thanks!

Rendered markdown link (for ease of readability): https://github.com/khushi-411/numpy.org/blob/master/content/en/benchmark.md

Future work which should be added here according to me are:

* [ ] **Algorithms:** Classification of Algorithms. Here, I classified it into 5 categories (Transonic jit, Transonic boost, Numba, NumPy, Pure NumPy). Future aims will be to update this into Cython, Pythran, Numba, TransonicJit, Transonic boost, Transonic Cython is any other relevant algorithms that I'm presently not sure about.
* [ ] **Graphs:** Do we need to iterate over different datasets? We can just select any one or two datasets and the number of times those algorithms are to be implemented and then plot the graph. Does it sound viable?
* [ ] **Content:** Improve Documentation (Include introduction). I mentioned various libraries in the post. Is that needed? I think that it can be presented in a better way. What do you think about how can I improve it?
* [ ] **Code Quality:** Profile code and optimize it.

cc: @mattip


content/en/benchmark.md
```diff
@@ -0,0 +1,128 @@
...
+  <td>Pure NumPy</td>
+  <td><a href = "https://github.com/khushi-411/numpy-benchmarks/blob/0.3/python/bench_numpy_highlevel.py">bench_pure_numpy_highlevel.py</a></td>
+</tr>
+  </table>
+</html>
+
+
+## Conclusion
+
+* These benchmarks are run with an Intel Core i7 processor.
```
rgommers, line 122: This probably belongs in the caption of the graph. With a bit more details about the exact CPU. For example:
```
$ cat /proc/cpuinfo | grep Intel
```
And then slightly reformat like: `Intel i9-7920X @ 2.90GHz`

content/en/benchmark.md
```diff
@@ -0,0 +1,128 @@
...
+  <td>169.4</td>
+ </tr>
+</table>
+</html>
+
+## Code Folders
+
+* These codes are originally taken from [here](https://github.com/paugier/nbabel)
+* Benchmark code: [here](https://github.com/khushi-411/numpy-benchmarks/blob/0.3/python/benchmark-2.py)
+* Graph: [here](https://github.com/khushi-411/numpy-benchmarks/blob/0.3/python/plot-modified-1.py)
```
rgommers, line 88: Linking to your own repo is fine for now. We can make it a separate NumPy repo or put the scripts in an existing repo somewhere, when you are closer to done with this PR.

content/en/benchmark.md
```diff
@@ -0,0 +1,128 @@
...
+![output-modify-2-new](https://user-images.githubusercontent.com/62256509/127653679-5151e22d-d141-4621-abb9-b3e6dd7988b7.jpg)
+
+
+## Implemented Algorithms
+
+Till this point 5 algorithms are implemented as mentioned bellow:
+
+### NumPy
+
+NumPy is a most fundamental package for scientific computing in Python. The major advantage of using NumPy is these gives same computational speed as C and Fortran. Let's see our inplementation details:
```
rgommers, line 13: I'm thinking what should come next is indeed a comparison with C/C++ and Fortran, like this sentence says?

The accelerators can be a second part of the story.

content/en/benchmark.md
```diff
@@ -0,0 +1,128 @@
+# NumPy Benchmarking
+
+
+![output-modify-2-new](https://user-images.githubusercontent.com/62256509/127653679-5151e22d-d141-4621-abb9-b3e6dd7988b7.jpg)
```
rgommers, line 4: It would be better to commit the figure in this PR. The place to put it is `static/images/content_images/`

rgommers (commented):
Thanks for working on this @khushi-411! Looks like a good start.

content/en/benchmark.md
mattip: The text should explain that this is a numerical simulation of the N-body problem and giv ethe constraints of the problem

C/C++ and Fortran code simulations can be found in the original repo https://github.com/paugier/nbabel/

content/en/benchmark.md
khushi-411: Apologize, I will update that. Thanks!

content/en/benchmark.md
khushi-411: Sure definitely, these are just sample codes. I feel that these need to be improved, and once it's finalized, we can create a repo or update somewhere in the existing repo as you'll wish. Thanks!

PR #463, opened 2021-08-31
Added news item about CZI DEI grant

melissawm:

Includes link to the full proposal and relevant Medium/press release links.

cc @rgommers


rgommers (commented):
This text looks great, thanks @melissawm! One other thought: how about updating the news bar on the front page for this as well? If you agree, it's in `content/en/config.yaml`, line 19-22.

melissawm:
Let me know if this is good - not sure how long we want this text to be!

rgommers (approved):
That looks good to me, thanks!

PR #467, opened 2021-09-21
Update link to mailing list archive from Nabble to Mailman 3

rgommers:


rgommers:
All green, so merging.

Issue #138, opened 2020-01-09
Add license text or a link to it to new website


InessaPawson:
@rgommers Would you like to add the content of this [page](github.com/numpy/numpy.org/blob/master/old-content/license.rst) to the new website?

rgommers:
Thanks for the suggestion Inessa. No, I don't want a license page - I much prefer to link to the license file on GitHub, that avoids maintaining duplicate content. 

I'll leave this issue open till an appropriate link is put in.

shaloo:
Taking a shot at this one as suggested by @rgommers.  Using https://github.com/numpy/numpy/blob/master/LICENSE.txt in and linking license info mentioned in 'Open Source' section under 'In a Nutshell'.

PR #471, opened 2021-10-02
Release website under BSD license

jarrodmillman:
I couldn't find what license the website content and theme was released under.  If I missed it, please let me know.  If not, this PR adds a revised BSD license to the repo.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2010-2021, NumPy Developers
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2020-2021, NumPy Developers
```
rgommers, line 3: Let's update this to 2010, that's the oldest commit in this repo (just not in the `master` branch). The new site started in 2019.

rgommers (commented):
Thanks @jarrodmillman, good idea. This repo is indeed missing a separate license file. The website itself links to https://github.com/numpy/numpy/blob/master/LICENSE.txt (see `content/<language>/about.md`), can you change that to point to this new license file?

LICENSE
jarrodmillman: Done.

jarrodmillman:
I am happy to do whatever, but I want to double-check whether I should update `content/<language>/about.md`.  That seems to be a reference to the license for the numpy package and not numpy.org.  Are you sure I should update that to point to the numpy.org LICENSE file?

Here is an old discussion about this:
https://github.com/numpy/numpy.org/issues/138

rgommers:
>  That seems to be a reference to the license for the numpy package and not numpy.org. Are you sure I should update that to point to the numpy.org LICENSE file?

Ah you are right! Okay then this is good to go.

rgommers (approved):

PR #472, opened 2021-10-02
Update default NumPy branch

jarrodmillman:
This replaces
- `github.com/numpy/numpy/tree/master/` with `github.com/numpy/numpy/tree/main/`

and

- `github.com/numpy/numpy/blob/master/` with `github.com/numpy/numpy/blob/main/`

This reflects the change from `master` to `main` as the name of the default branch.

### DIFF
--- a/content/en/about.md
+++ b/content/en/about.md
@@ -5,7 +5,7 @@ sidebar: false
 
 _Some information about the NumPy project and community_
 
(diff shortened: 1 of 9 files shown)


rgommers (changes requested):
Thanks @jarrodmillman. We cannot merge this now though, see my comment on https://github.com/numpy/numpy.org/issues/473#issuecomment-932789706.

I would prefer if we closed this PR for the time being, and linked it from gh-473 so we can revive it at a time when we're either already making necessary deployment changes or when I have significantly more bandwidth.

rgommers:
O wait, I connect it in my head to gh-473, but this is different - just fixing links. Sorry for the noise, let me check again.

rgommers (approved):
Never mind my previous comment, this just fixed links - thanks @jarrodmillman!

PR #500, opened 2021-11-16
Update theme

jarrodmillman:
Moving more stuff back from theme.  Everything should be the same except for the news.

### DIFF
--- a/content/en/config.yaml
+++ b/content/en/config.yaml
@@ -15,11 +15,6 @@ params:
     buttonlink: "/install"
     # Hero image (from static/images/___)
     image: logo.svg
-  # Customizable navbar. For a dropdown, add a "sublinks" list.
-  news:
-    title: D&I Grant from CZI
-    content: Including NumPy, SciPy, Matplotlib and Pandas
-    url: /news
 
   shell:
     title: placeholder
--- a/content/en/news.md
+++ b/content/en/news.md
@@ -1,7 +1,10 @@
 ---
 title: News
 sidebar: false
+newsHeader: D&I Grant from CZI
+date: 2021-08-31
 ---
+
 ### Advancing an inclusive culture in the scientific Python ecosystem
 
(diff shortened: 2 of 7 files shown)


stefanv:
Current news:



New news (title followed by date instead of phrase):

melissawm:
I am ok with this, maybe @InessaPawson and @MarsBarLee should also take a look.

melissawm (approved):

InessaPawson:
I’d rather leave dates out. We don’t share news often, but the ones that make it on the news page are of great importance. By placing dates in the news bar we will be drawing attention to the information that is pretty trivial for the most stakeholders.

stefanv:
I must say, to my eye the current version reads quite oddly. A date associated with a news item feels very natural! If you look at the SciPy news releases, e.g., they all start with a date.

Let's see what the others say too.

InessaPawson:
To support my preference, see https://kotlinlang.org. 🤓

stefanv:
I don't follow: their site does not seem to have a news/announcement section?

InessaPawson:
@stefanv It's at the very top of the page (see the screenshot below).

Kotlin_newsbar

stefanv:
Interesting, I don't see that banner in my browser!

stefanv:
FWIW, I think it's OK to have a news announcement without any side-text.  But if there is side-text, a date seems appropriate.  The date can be suppressed in our current theme by simply not adding a date to the news site (it may be that I have to tweak that to work, but that'd be easy enough).

InessaPawson:
No side-text would be fine. (I frankly don’t remember why we have it.) Thank you for being flexible!
P.S. The current announcement in the news bar could benefit from some tweaking.

jarrodmillman:
@InessaPawson Sorry for the delay.  We've updated the theme to remove the side-text (if the date is not specified).  See 194e5d8ec5aa.  Now the news banner looks like this:
screenshot

InessaPawson (approved):
This looks great! Thank you, @jarrodmillman and @stefanv!

rgommers (approved):
LGTM too, thanks @jarrodmillman, all!

PR #504, opened 2021-12-10
Announce the release of NumPy 1.22.0rc2.

charris:


rgommers (approved):
Thanks Chuck. Let's get this merged. I think we should delete the entries again once we add `1.22.0`. In the future, we may want to skip these changes for rc's completely.

PR #510, opened 2022-01-03
New gallery

jarrodmillman:
This switches to using the scientific-python-hugo-theme's new gallery style (scipy and scientific-python are already using it).  This uses the NumPy GitHub organization as the definitive list of team members.  So we may want to make sure the organization teams are correct before merging this.  There are also some missing teams (i.e., `documentation` and `funding and grants`).

There are instructions on how to update the galleries in the `README.md`.  But basically, you need to

1. Generate a personal GH token with read access to organizations
2. Export that token as `GH_TOKEN`
3. Run `make teams`
4. git add, git commit, git push
5. make a PR and have it merged


jarrodmillman:
I tried translating the `teams.md` page using Google translate.  How are translations handled?

Makefile
```diff
@@ -21,6 +21,23 @@ prepare:
...
+
+$(TEAMS_DIR)/%.md: $(TEAMS_DIR)
+	$(TEAMS_QUERY) --org numpy --team "$*"  >  $(TEAMS_DIR)/$*.html
+
+teams-clean: prepare
+	for team in $(TEAMS); do \
+	  rm -f $(TEAMS_DIR)/$${team}.html ;\
+	done
+
+teams: | teams-clean $(patsubst %,$(TEAMS_DIR)/%.md,$(TEAMS))
```
stefanv, line 39: I realized recently that piped operators do not always behave the way I thought they did. Specifically, I think everything after the pipe is ordered (what I wanted) but may be evaluated as phony targets (not what I wanted). Could you please double check?

stefanv (approved):
I can't speak to the translations, but everything else looks good.

jarrodmillman:
@stefanv I did https://github.com/numpy/numpy.org/pull/510/commits/4e25210313411f9c7b47623d2f50c6cda076ae37 b/c the links were using numpy.org as the base URL on the deploy builds.  That wasn't happening for me locally.  Do you have any idea how to fix it so we can use Hugo links?  Or do you think what I did is reasonable?

rgommers:
Everything in `.md` and `.yml` files gets imported into Crowdin for translations. So you don't need to worry about it when adding new content - that would not work at all when the number of languages grows.

It's probably best to revert the machine translations (other than fixed links). Then they can go through the regular translation/review process. The one word of Portuguese should be fine, but I'd prefer not to merge Japanese characters unless a native speaker has looked at them.

jarrodmillman:
@rgommers Should I just copy `content/en/teams.md ` to `content/ja/teams.md` and `content/pt/teams.md`?  If those files are missing, I get the following
```
8:23:13 AM: ERROR 2022/01/03 16:23:13 [pt] REF_NOT_FOUND: Ref "/teams": "/opt/build/repo/content/pt/about.md:46:27": page not found
8:23:13 AM: ERROR 2022/01/03 16:23:13 [ja] REF_NOT_FOUND: Ref "/teams": "/opt/build/repo/content/ja/about.md:46:62": page not found
```

rgommers:
Yes, just copying sounds good.

InessaPawson (commented):
Thank you for taking care of it, @jarrodmillman!
I’ve pinged everyone to update their teams. (I don’t have permissions.)
As far as I know, we haven’t established an official Funding/Grants team yet.

rgommers:
> I’ve pinged everyone to update their teams. (I don’t have permissions.)

I looked at this, and I think we need some guidelines as to who is on a team and what that means. I meant to write a NEP about this, but didn't get around to that. Maybe that's something we can do together.

+1 to adding translators separately, that seems like a good idea (can be a separate PR later).

rgommers:
It stands out for the Triage team in particular that the team is huge, but almost no one on that team is doing any actual triaging. It seems like we need two levels: "team member" (which are people who take some amount of ownership and are actively involved), and "has permissions on some repo(s)". Triage team is all the latter right now.

rgommers:
I think it's fine to merge this in its current form and revisit that topic, because it is going to take a while to sort out. Any objections to that?

InessaPawson:
Formalizing NumPy teams via NEP is a good idea. Let's merge this as it is and try to update the rest within the next 4 weeks.

stefanv:
> @stefanv I did [4e25210](https://github.com/numpy/numpy.org/commit/4e25210313411f9c7b47623d2f50c6cda076ae37) b/c the links were using numpy.org as the base URL on the deploy builds. That wasn't happening for me locally. Do you have any idea how to fix it so we can use Hugo links? Or do you think what I did is reasonable?

I think it's best to use [relref](https://gohugo.io/functions/relref/).

jarrodmillman:
Before we merge, I have a couple minor cleanups that I will take care of later today.

Makefile
jarrodmillman: I marked `teams` and `teams-clean` as a phony target in bb321af.  So now they both always execute regardless of how piped operators behave.

jarrodmillman:
I finished making my changes and this is ready to merge.

stefanv:
Now that this is in, I wonder if we should move the one level of indirection to the gallery. Is there a reason not to put it directly under the "About Us" page, since we already list teams and steering council members there? We can probably add a steering council GH group and render that just like we do the others.

InessaPawson:
@stefanv To clarify, are you suggesting to add the content of https://numpy.org/teams/ directly to the About Us page (https://numpy.org/about/)?

stefanv:
Yes, exactly, or move the steering council onto the teams page.  But I have a preference for the first option---less clicking.

InessaPawson:
Your suggestion makes a lot of sense as there is a noticeable overlap of content between the two pages.

PR #526, opened 2022-01-06
Print Makefile help via Python; reorganize targets

stefanv:


jarrodmillman (approved):

charris:
@stefanv Do you want to pursue this?

stefanv:
I can fix it up, if my PR is causing CI to fail. I'm on leave now, but can get to it after I return.

jarrodmillman:
@alphapapa this is the Makefile PR I mentioned.

InessaPawson:
@stefanv Given the [latest discussion on this topic](https://github.com/numpy/numpy.org/issues/680), do you still want to implement this?

scripts/makefile_to_help.py
```diff
@@ -0,0 +1,31 @@
...
+
+Usage: makefile_to_help.py <MAKEFILE>
+
+"""
+
+import sys
+import re
+
+
+if not sys.argv:
```
shoumikchakravarty-dev, line 14: sys.argv always contains at least the script name (sys.argv[0]), so this guard is never triggered.  If the script is
called with no argument, it falls through to sys.argv[1] on the last line and crashes with an IndexError instead of printing the usage message.

Suggested fix 
 if len(sys.argv) < 2:
      print("Usage: makefile_to_help.py <MAKEFILE>")
      sys.exit(1)  # exit code 1 for error, not 0

scripts/makefile_to_help.py
```diff
@@ -0,0 +1,31 @@
...
+    targets = {target: desc for (target, desc) in groups}
+
+    N = max(len(target) for (target, desc) in targets.items())
+
+    for (target, desc) in targets.items():
+        print(f'{target:{N}} {desc}')
+
+
+fname = sys.argv[1]
+describe_targets(open(fname).readlines())
```
shoumikchakravarty-dev, line 31: The file handle is never explicitly closed. Use a with block.
 with open(fname) as f:
      describe_targets(f.readlines())

scripts/makefile_to_help.py
```diff
@@ -0,0 +1,31 @@
...
+    groups = [m.groups(0) for m in matches if m]
+    targets = {target: desc for (target, desc) in groups}
+
+    N = max(len(target) for (target, desc) in targets.items())
+
+    for (target, desc) in targets.items():
+        print(f'{target:{N}} {desc}')
+
+
+fname = sys.argv[1]
```
shoumikchakravarty-dev, line 30: If sys.argv[1] is a path that doesn't exist, Python raises an unguarded FileNotFoundError. A simple check would give a cleaner error message.

shoumikchakravarty-dev (commented):
Pls review the comments added

PR #546, opened 2022-02-01
Use plausible analytics

jarrodmillman:
Trying plausible analytics.  We are using this for scipy.org, docs.scipy.org, and scientific-python.org.  If this PR is merged, you will be able to view the analytics for numpy.org at https://views.scientific-python.org/numpy.org.  The page is currently private, so you will need to create a plausible account and let @stefanv or me know what address you registered.

I don't know who currently has access to the google analytics, but we should make sure whoever is currently monitoring the google analytics is OK with using plausible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,3 +79,9 @@ To update the teams gallery in numpy.org site, you need to run `make teams` Make
 export GH_TOKEN=xxxxxxxxxx
 make teams
 ```
+
+## Analytics
+
+A self-hosted version of [Plausible.io](https://plausible.io) is used to gather simple
+and privacy-friendly analytics for the site. The dashboard can be accessed
+[here](https://views.scientific-python.org/numpy.org).
--- a/config.yaml.in
+++ b/config.yaml.in
@@ -21,6 +21,8 @@ params:
   font:
     name: "Lato"
     sizes: [400,900]
+  plausible:
+    dataDomain: numpy.org
 languages:
   en:
     title: NumPy


rgommers:
Thanks Jarrod! Could I get an invite?

More importantly, I don't think it's working well, these numbers are definitely way too low, especially for the docs (should be O(100,000/day):

image

stefanv:
Two things: (1) we currently only track `/devdocs` and (2) Plausible does respect AdBlock, which takes down a further fraction.

rgommers:
Hmm yeah, I'd expect many/most people to use AdBlock (how can you tolerance browsing without it?). Kind of a strange decision, because this has nothing to do with ads. Any way to change that behavior?

Statistics should improve when we have Plausible integration in a release rather than only in `devdocs`, but it's still not great to throw away so much signal. 

For the record, this is what Cloudflare says for `numpy.org` right now:

image

stefanv:
We can host the plausible script on scientific-python.org, which should be unlikely to be blocked (https://plausible.io/docs/proxy/introduction#are-you-concerned-about-missing-data).  I can set that up tomorrow.

rgommers:
that sounds good, thanks

stefanv:
Looks like we're already doing that, so I'll have to check why adblock picks it up; could be the naming of the script.

stefanv:
EasyList contains:

```
/plausible.js$domain=~plausible.io
/plausible.outbound-links.js$domain=~plausible.io
```

So will have to rewrite the JavaScript name.

InessaPawson:
@jarrodmillman @stefanv First of all, thank you for finding an open source alternative to GA! 

As someone who used to spend a lot of time on this type of analytics, I would be curious to compare results from both platforms for the next 2-3 months.

stefanv:
I've implemented a fix for the ad-blocking issue on analytics.scientific-python.org and in the hugo theme.

stefanv:
The analytics server had to be renamed to avoid uBlock, so now it is located at https://views.scientific-python.org.  I also added a rewrite rule to serve the Javascript from a more innocuous location, and the hugo theme has been updated accordingly. This should take care of most adblock issues.

README.md
```diff
@@ -79,3 +79,9 @@ To update the teams gallery in numpy.org site, you need to run `make teams` Make
 export GH_TOKEN=xxxxxxxxxx
 make teams
 ```
+
+## Analytics
+
+A self-hosted version of [Plausible.io](https://plausible.io) is used to gather simple
+and privacy-friendly analytics for the site. The dashboard can be accessed
+[here](https://analytics.scientific-python.org/numpy.org).
```
stefanv, line 87:
```suggestion
[here](https://views.scientific-python.org/numpy.org).
```

InessaPawson (commented):
The PR looks good to me. I’m just thinking, maybe, we should keep Google Analytics for another month or two. In the past, regardless of the analytics platform, it took at least a month for new accounts to get more reliable data on website traffic.

jarrodmillman:
I added back google analytics.  Since the theme no longer supports this, I added the necessary files to this repo (see #549).  If this PR is merged, I will create an issue to make sure we don't forget to clean up this "temporary" fix when (or if) numpy.org decides to stop using google analytics.

@InessaPawson Do you have access to the google analytics (I don't)?  If not, do you know who does?

jarrodmillman:
I decided to merge this as it doesn't remove anything, everyone is interested in trying it out, and we need to merge it to compare it with google analytics.  If this causes any issues, I will fix them or undo this change.  To make sure we don't loose track, I created #550 where we can continue the discussion.

InessaPawson:
> I added back google analytics. Since the theme no longer supports this, I added the necessary files to this repo (see #549). If this PR is merged, I will create an issue to make sure we don't forget to clean up this "temporary" fix when (or if) numpy.org decides to stop using google analytics.

Sounds great! 

> @InessaPawson Do you have access to the google analytics (I don't)? If not, do you know who does?

Yes, I do. @rgommers is another web team member who has it.

InessaPawson:
@jarrodmillman May I have an invite to the numpy.org plausible account?

PR #553, opened 2022-02-03
Update theme (shortcuts improvements)

jarrodmillman:
The shortcuts/sidebar has been fixed. It now doesn't scroll up with the rest of the page and the links work. It also makes the appearance more like what you see with pydata-sphinx-theme.

## Old look

screenshot

## New look


stefanv:
scrolly-shortcuts

stefanv:
I should note that only `h` headers with ids will appear on the sidebar, and I see some of the NumPy headings don't; so that may be something else to fix in this PR, @jarrodmillman.

jarrodmillman:
I would rather keep the style and content changes separate.  Maybe someone could fix the headings in a separate PR.

stefanv:
That works!

jarrodmillman:
Just to be clear, the headings issue @stefanv mentioned is unrelated to this PR.  The headings haven't changed.

PR #556, opened 2022-02-10
Cleanup unused code

### LINKED ISSUE #555: Clean up unused code and assets
This is a follow-up issue to https://github.com/numpy/numpy.org/pull/547.

Now that the website uses an embeddable JupyterLite code console, there might be some unused code, assets and CSS still around that can be cleaned up.

jtpio:

Fixes #555 

Follow-up to #547 

Also tweak the "Try NumPy" title so it's centered:

image


stefanv (approved):

jarrodmillman:
This looks great!  One minor nit, would it be possible to add the gray background to the `Try NumPy` heading / text
screenshot

Just to make it clear that the `Try NumPy`  is part of the gray box below it.

stefanv:
One thing I noticed: matplotlib works in these cells, but not until you've typed `%matplotlib inline`. Is there a way to set up the notebook so that matplotlib works out of the box?

jtpio:
> This looks great! One minor nit, would it be possible to add the gray background to the Try NumPy heading / text

Done in https://github.com/numpy/numpy.org/pull/556/commits/812928a10679e0cf7b76727d1545b47758068341

jtpio:
> but not until you've typed `%matplotlib inline`. Is there a way to set up the notebook so that matplotlib works out of the box?

This seems fine when using the following code snippet?

```python
import numpy as np
import matplotlib.pyplot as plt

x = np.linspace(0, 10, 1000)
plt.plot(x, np.sin(x));

plt.show()
```

image

stefanv:
Ah, okay: I guess no one uses plt.show in notebooks, so I didn't even think to try it.

jobovy:
@stefanv: Once https://github.com/jtpio/replite/pull/24 is merged, it would be possible to pre-run `%matplotlib inline` so users wouldn't have to do `plt.show()`. I would think that's better, although in my documentation experience, users have found it confusing when `%matplotlib inline` is silently assumed for code examples (lots of emails asking why plots don't show up for them from novice Python users). So another option would be to have it be part of the first `import numpy as np` cell (but that might lead to its own confusion...)

stefanv:
@jobovy Yes, explicit is probably better. We could, e.g., add a tab to the existing NumPy example snippet, so that we have multiple snippets—one of which can show how to plot.

jtpio:
> e.g., add a tab to the existing NumPy example snippet

This sounds good.

At a later stage we could even consider having an (opt-in) console closer to the examples using `matplotlib`, for instance here: https://numpy.org/doc/stable/reference/generated/numpy.absolute.html:

image

jarrodmillman (approved):

stefanv:
Thanks @jtpio!

PR #557, opened 2022-02-15
Switch to the upstream JupyterLite REPL app

jtpio:

Follow-up to #547 and #556 

JupyterLite now ships with the `repl` app by default:

- Added in: https://github.com/jupyterlite/jupyterlite/pull/498
- Documentation: https://jupyterlite.readthedocs.io/en/latest/applications/repl.html
- Changelog: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.1.0a23

This PR updates the "Try NumPy" section:

- [x] The URL of the interactive code console now points to the JupyterLite demo repo: https:/jupyterlite.github.io/demo/repl/?toolbar=1&kernel=python&code=import%20numpy%20as%20np
- [x] Remove the "Enable" button, since updating to the upstream code console also fixes the scrolling issue initially reported in https://github.com/numpy/numpy.org/pull/547#issuecomment-1028686114
- [x] Clean up unused styles

https://user-images.githubusercontent.com/591645/154056572-4675b611-00c0-4488-b8b7-99d2ece07215.mp4

---

A follow-up to this would be to start making a custom JupyterLite deployment that can then be hosted in the `numpy` GitHub organization, for example at https://github.com/numpy/try or https://github.com/numpy/jupyterlite-repl.

The custom deployment can easily be done using this template repo: https://github.com/jupyterlite/demo

### DIFF
--- a/assets/css/shell.css
+++ b/assets/css/shell.css
@@ -48,28 +48,6 @@
     font-weight: bold;
 }
 
-.shell-enable-button {
-    position: relative;
-    margin: 15px 0;
-}
-
-.shell-button {
-    width: 225px;
-    height: 50px;
-    font-size: 12px;
-    text-transform: uppercase;
-    letter-spacing: 1.5px;
-    font-weight: 500;
-    color: rgb(0,0,0);
-    background-color: rgb(238, 238, 238);
-    border: none;
-    border-radius: 25px;
-    outline: none;
-    /* Black, with 10% opacity */
-    box-shadow: 0px 8px 15px rgba(0, 0, 0, 0.1);
-    cursor: pointer;
-}
-
 .shell-lesson {
     display: flex;
     text-align: left;
--- a/content/en/config.yaml
+++ b/content/en/config.yaml
@@ -21,14 +21,9 @@ params:
 
     intro:
     - title: Try NumPy
-      text: Enable the interactive shell
+      text: Use the interactive shell to try NumPy in the browser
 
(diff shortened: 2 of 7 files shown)


stefanv:
Thanks @jtpio! This looks good.

Can you expand on the advantages of a custom JupyterLite deployment?

I'm still a bit worried about the downloads here: at >15MB per notebook load, your jupyterlite.github.io server may be hit with much greater than the 100GB soft limits that GitHub has in place. So, may make sense to (a) load the terminal on demand and (b) put a cache such as Cloudflare in front of your pages.

jtpio:
> Can you expand on the advantages of a custom JupyterLite deployment?

A custom deployment gives more control on the version of JupyterLite being used, the configuration, custom settings and the set of extensions available.

And this could also help deploy it on Netlify alongside the rest of the website if desired.

> I'm still a bit worried about the downloads here: at >15MB per notebook load, your jupyterlite.github.io server may be hit with much greater than the 100GB soft limits that GitHub has in place. So, may make sense to (a) load the terminal on demand and (b) put a cache such as Cloudflare in front of your pages.

Normally the size of the assets downloaded from https://jupyterlite.github.io/demo should be a bit smaller than 15MB:

image

Also these should be cached by the browser on subsequent visits. Also part of the downloads includes Pyodide, which is fetched from `cdn.jsdelivr.net`.

Option a) might have the downside of reintroducing some waiting time if the assets and the kernel initialization is triggered on user action and not pre-fetched in the background?

stefanv:
Thanks @jtpio!

PR #567, opened 2022-03-16
Improve JupyterLite console alignment

SylvainCorlay:
This improves the alignment of the code console with respect to the example snipet

**Before:**

before

**After:**

after

### DIFF
--- a/assets/css/shell.css
+++ b/assets/css/shell.css
@@ -1,40 +1,27 @@
 .hero-right {
     display: flex;
     flex-direction: column;
-    width: 100vw;
     /* Black, with 10% opacity */
     box-shadow: 0px 8px 15px rgba(0, 0, 0, 0.1);
 }
 
 .numpy-shell-canvas {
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
cc @jtpio.

jtpio:
Nice, thanks!

Looks like the "Try Numpy" section is now taking the full width of the page:

image

Instead of being centered. Looks fine to me although other folks might have a different opinion.

SylvainCorlay:
OK, I can put a max width if people prefer that.

SylvainCorlay:
Improved things a bit by having a max width and the shell centered.

jtpio:
Looks good!

stefanv:
Looks great, thanks! Can we add back in a bit of bottom padding? I see it's all gone.

SylvainCorlay:
@stefanv done.

stefanv:
Thanks @SylvainCorlay!

PR #587, opened 2022-07-29
Update theme (footer icons)

### LINKED ISSUE #441: Increase the size of the social media logos and change the color  of them in the Footer
The size of Social Media logos in the Footer is too small that sometimes it is not even visible.
Also, the color used in the logos is matching with the Footer color so it is hard to notice the logos so use contrast colors in the logo other than Footer color so that it is clearly visible.

I'd love to work on this and solve this.
Please give me a green signal and assign me this task.

jarrodmillman:
Close #441. Thanks to @AsavariA .


jarrodmillman:
I am merging this as the change was requested by numpy.org and the theme PR was already approved. If this turns out to be an issue, we can revert the change.

iaiu2142:
Thanks for doing this @Asavari

PR #602, opened 2022-10-06
DOC: Update ordering details and some translations

HaoZeke:

~- Updated the order of names in the `en/about` page to be in alphabetical order (as promised by the page)~
- Clarified the ordering of names
- Updated the names and order of `{ja,pt}/about` pages

### DIFF
--- a/content/ja/about.md
+++ b/content/ja/about.md
@@ -14,24 +14,26 @@ NumPy は 、NumPyコミュニティやより広範な科学計算用Python コ�
 
 NumPy運営委員会の役割は、NumPyのコミュニティと協力しサポートすることを通じて、技術的にもコミュニティ的にも長期的にNumPyプロジェクトを良い状態に保つことです。 NumPy運営委員会は現在以下のメンバーで構成されています (アルファベット順):
 
-- Sebastian Berg
-- Jaime Fernández del Río
-- Ralf Gommers
-- Allan Haldane
 - Charles Harris
-- Stephan Hoyer
-- Matti Picus
-- Nathaniel Smith
-- Julian Taylor
-- Pauli Virtanen
-- Stéfan van der Walt
 - Eric Wieser
+- Inessa Pawson
+- Matti Picus
+- Melissa Weber Mendonça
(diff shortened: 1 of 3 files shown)


stefanv:
The old list was alphabetical by last name, which made sense. The new list is... not alphabetical :)

stefanv:
Last name ordering makes more sense to me; I recommend we leave as-is.

HaoZeke:
> Last name ordering makes more sense to me; I recommend we leave as-is.

Oops, updated the document for less confusion for me (and maybe others); also updated the translations.

stefanv (approved):
Alphabetizing last names is hard! There should be one-- and preferably only one --obvious way to do it.
Although that way may not be obvious at first unless you're Dutch.

Thanks for tidying up!

content/ja/about.md
```diff
@@ -12,26 +12,28 @@ NumPy は 、NumPyコミュニティやより広範な科学計算用Python コ
...
-NumPy運営委員会の役割は、NumPyのコミュニティと協力しサポートすることを通じて、技術的にもコミュニティ的にも長期的にNumPyプロジェクトを良い状態に保つことです。 NumPy運営委員会は現在以下のメンバーで構成されています (アルファベット順):
+NumPy運営委員会の役割は、NumPyのコミュニティと協力しサポートすることを通じて、技術的にもコミュニティ的にも長期的にNumPyプロジェクトを良い状態に保つことです。 NumPy運営委員会は現在以下のメンバーで構成されています (アルファベット順、姓で):
 
 - Sebastian Berg
-- Jaime Fernández del Río
 - Ralf Gommers
-- Allan Haldane
 - Charles Harris
 - Stephan Hoyer
+- Melissa Weber Mendonça
```
melissawm, line 21: Well, there's the issue... Weber is a last name (my mother's side) so technically I should go after Stéfan I think? 

Otherwise, looks good :)

HaoZeke: Excellent point, moved things around ^_^

stefanv:
Thanks!

Issue #478, opened 2021-10-15
Change icon colors in Science tab of Ecosystem section


MarsBarLee:
Change the colors of the icons in the Science tab of the Ecosystem section from black to the dark blue used in the rest of the theme, hex color code #013243.

Related to #474. This issue is to help separate the multiple issues mentioned in #474 into their own separate issues. @new-vivek11, would you like to continue with making the color changes?

MarsBarLee:
@new-vivek11 You make a contribution by [making a pull request](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) in this [numpy.org repository](https://github.com/numpy/numpy.org/pulls).

If you're not already familiar with using Git on the command line, you can also make a pull request (PR) using the Github.com. 

You can make a fork of numpy.org repository. Then go to your fork and make a new branch. On that branch, you can replace the files with the new ones by deleting the old ones and uploading the new ones (there should be buttons for both 'delete' and 'add files'). Then you make a pull request from your branch to this [repository](https://github.com/numpy/numpy.org/pulls). This will open a new pull request page.

By making your own PR, you get contribution recognition to your account! :)

new-vivek11:
i have made a pull request : https://github.com/numpy/numpy.org/pull/479

not sure whether i did it the right way or not but my guess is that i have done it right (new to pull requests) ... please let me know whether i have done all that i need to do from my side. @MarsBarLee

PR #607, opened 2022-10-27
Change icon colors in Sience Domains tab of Ecosystem section

ErnestoFGonzalez:

- Fixes gh-478.
- Set icon colors in Science domain tab of Ecosystem section to `#013243`.

Following is a screenshot of the result:

Screenshot 2022-10-28 at 00 17 43


stefanv:
If we're updating the icon colors, should we update the headings to match? Also, the icons are currently centered which looks a bit weird: how about left alignment, as with the headings?

ErnestoFGonzalez:
> If we're updating the icon colors, should we update the headings to match? Also, the icons are currently centered which looks a bit weird: how about left alignment, as with the headings?

Yeah, that'd look better. I'll push another commit.

ErnestoFGonzalez:
I like it better this way. What do you think @stefanv?

stefanv:
That looks good! I see `casestudies.css` also custom specifies color. I recommend refactoring that into a variable that is reused elsewhere.

Actually, looking at the theme CSS you already have:

```
  --colorPrimaryDark:  rgb(1, 50, 67);
```

which you can reuse as:

```
color: var(--colorPrimaryDark);
```

ErnestoFGonzalez:
> That looks good! I see `casestudies.css` also custom specifies color. I recommend refactoring that into a variable that is reused elsewhere.
> 
> Actually, looking at the theme CSS you already have:
> 
> ```
>   --colorPrimaryDark:  rgb(1, 50, 67);
> ```
> 
> which you can reuse as:
> 
> ```
> color: var(--colorPrimaryDark);
> ```

Right! I missed that. 
Done for `case studies.css` and `tabs.css`. 
I see some usages of the hex color in `layouts/shortcodes/numfocus.html`. Want me to see if I can use the variable here?

stefanv (approved):

stefanv:
Don't worry about the NF partial, I think this is good to go!

PR #608, opened 2022-10-29
Align images for multiline titles in CS section

### LINKED ISSUE #416: Alignment of titles in the (translated) Case Studies section
When the titles for some of the Case Studies are translated (see #55), they may take more than one line of text, which results in the images being misaligned:

Screenshot_20210419_172809

ernestofgonzalez:
Fixes gh-416

1. Set images width and height to be the same. The `static/images/content_images/case_studies/deeplabcut.png` has different aspect ratio than the rest, which resulted in misalignment with other cards in 2x2 grid.
2. Set title height to be the double of line height. This guarantees alignment of all images in case of titles with two lines.

A note on point 2: A more interesting approach could be to check the number of lines of the longest title and set the line height of all titles to the line height of that one. I tried this but couldn't make the height of the card to update accordingly.

### DIFF
--- a/assets/css/casestudies.css
+++ b/assets/css/casestudies.css
@@ -38,6 +38,12 @@
     text-transform: uppercase;
 }
 
+.casestudies-box-content img {
+    object-fit: cover;
+    width: 350px;
+    height: 218.5px;
+}
+
 .casestudies-box-text {
     margin: 30px 15px;
     font-size: 14px;
--- a/assets/css/casestudies.css
+++ b/assets/css/casestudies.css
@@ -47,7 +47,7 @@
 }
 
 .casestudies-box-text {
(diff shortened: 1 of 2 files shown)


ernestofgonzalez (commented):
The auto margin bottom was from another experiment I was running. I had forgot to remove it.

InessaPawson (commented):
Thanks for working on this and for your patience with review, @ernestofgonzalez ! The issue is still present, it would be great to address it.
One concern about the proposed solution is the `fixed height: 3em` for `.casestudies-box-title`. This assumes translated titles will never wrap beyond two lines. A longer translation could overflow into the image. Could we make this a little more robust?
Also, since the PR was submitted a while back, please update the branch.

PR #617, opened 2022-12-26
announce the NumPy 1.24.1 release

charris:


melissawm:
Can't see why this is failing but should be safe to merge.

Issue #489, opened 2021-10-26
Logo size of icons in array libraries tab


shaloo:
This is probably similar to #441 (unless I missed an issue to rework array libraries page).  Please resize to a standard size - all the logos in the first column on array libraries page. 
<img width="1008" alt="Screenshot 2021-10-26 at 3 05 54 PM" src="https://user-images.githubusercontent.com/5890484/138852077-92a56d16-2307-4e55-88fb-50b78dde876f.png">

rgommers:
Agreed it doesn't look good, thanks for the report @shaloo. This is in general a hard thing to do right, because the aspect ratios of these logos are wildly different. We could do some resizing; PyData Sparse is in need of a better logo with a width larger than the height.

MattiaAldighieri:
I'll try to work on this.

PR #621, opened 2023-02-04
Standard size of logos in array libraries tab

MattiaAldighieri:
Fixes [gh-489](https://github.com/numpy/numpy.org/issues/489#issue-1036065739)

The logos on the array libraries tab have different sizes, here they have a standard size (that might be changed, for example if reduced logos would be even more uniform).

Before:
image
After:
image


InessaPawson:
@MattiaAldighieri, this is definitely an improvement! Let's merge it.

PR #623, opened 2023-02-05
announce the NumPy 1.24.2 release

charris:


stefanv:
Thanks, Chuck! Would you like this to show up on the frontpage as well?

stefanv:
OK, I see, just the list of releases was updated. Is that what you want?

charris:
>Is that what you want?

Yep. We only do the extras for the first release in a cycle, and sometimes @rgommers edits things. If folks want more, they should mention it in the NumPy issues.

stefanv:
:+1:

PR #628, opened 2023-02-16
Add 'News' button to the top navigation bar

InessaPawson:


stefanv:
I don't know if this change requires further discussion, but it LGTM.

InessaPawson:
> I don't know if this change requires further discussion, but it LGTM.

We discussed this at today’s community call.

stefanv:
Then in it goes!

Issue #55, opened 2019-10-18
Multilingual support with Crowdin


InessaPawson:
For the up-to-date information on this issue refer to the following:
 https://github.com/Shekharrajak/scipy-sphinx-theme-v2/issues

Testing with the current NumPy doc: https://github.com/numpy/numpy/pull/14648

The latest Sphinx theme: [Demo link](https://pr-29-scipy-sphinx-theme-v2.surge.sh/)

Shekharrajak:
Refer: https://github.com/numpy/numpy.org/pull/61, https://github.com/numpy/numpy.org/pull/62

rgommers:
Just did some testing, this looks nice! I think for testing we can do Mandarin (@zjpoh offered to help) or Hindi if you want to do that @Shekharrajak. I think Dutch was good to try the interface, but we don't really need/want a Dutch translation - Dutch people are perfectly happy with English usually.

Shekharrajak:
Yeah! I will try to do that.

rgommers:
Related gh-329 for Korean translation.

rgommers:
I think we're good to go on this, the content is stable and everything seems to work. I'll re-enable the language selector as well.

@zjpoh would you still be interested to work on the Chinese translation? 

I've added a wiki page at https://github.com/numpy/numpy/wiki/Translations-of-the-NumPy-website on how to get started.

rgommers:
Re-enabling the dropdown is a matter of undoing the small patch in gh-171

rubentl134:
I would like to participate on the Spanish translation. Is there any way I can start? I feel kind of lost here. Sorry

rgommers:
That's great, thanks @rubentl134! Spanish is already enabled on Crowdin, so to get started see the instructions at https://github.com/numpy/numpy/wiki/Translations-of-the-NumPy-website. Basically, sign up for Crowdin via the link there, select Spanish, and start:)

If you have any issues, please let us know.

comradekingu:
Can you use Weblate instead, so I can contribute without getting spied on, leak info, and also contribute to their closed translation memory? You need a license to be able to use the translations btw. :)

rgommers:
> Can you use Weblate instead, so I can contribute without getting spied on, leak info, and also contribute to their closed translation memory? 

We have spent significant time setting up Crowdin, after good experiences Spyder had with it. And the Crowdin team has been quite helpful, so I see no reason to switch.

[...]

PR #647, opened 2023-06-20
Make internal link work for translated content

steppi:
This PR adds an anchor before the **Reproducible installs** heading in `install.md`. This will allow the internal link to this section work for translated content.

See https://github.com/numpy/numpy.org/issues/55#issuecomment-1599175991

Previously, anchors were added just to the translated content in gh-540, but when translations need to be updated, these will be stripped by Crowdin if the anchor does not appear in the English content.


rgommers (approved):
LGTM, thanks @steppi

PR #649, opened 2023-06-27
Update hugo to 0.115.4

jarrodmillman:
**DO NOT MERGE UNTIL UPDATED TRANSLATIONS ARE MERGED**

Depends on #639. See  https://github.com/numpy/numpy.org/pull/649#pullrequestreview-1502852081.

Updating hugo in preparation for updating the theme. There were some changes in hugo 0.112 that are causing issues (addressed in e749f62) : https://gohugo.io/content-management/multilingual/#changes-in-hugo-01120


rgommers (commented):
Thanks @jarrodmillman. These changes all seem useful, but I'd like to make sure that they won't cause problems for the imminent launch of the translated version of the site. We now have fully complete translations, and it seems to me that this may cause significant rework (not sure, I'm hoping @steppi can tell).

If there is a conflict, I'd much prefer to launch the Japanese and Portuguese translations first (should happen very soon), and then rebase this PR on top.

steppi:
I hadn't attempted to build this PR branch locally until just now. I made sure hugo 0.114.1 was installed. The rendered page only includes a top bar with a link to the news page. I've just checked the deploy preview. The same issue appears there as well.

image

stefanv:
Hugo 0.114 is forcing us to list non-standard parameters for each language under a `params` section. If the rendering is incomplete, it sounds like we also need to update the theme to pick up those values from a different location.

I think we'll go ahead and fix the theme for now, and then when NumPy is ready (after the language merge?) we can just move the relevant sections under `lang: params: ...` and update the theme.

stefanv:
@steppi I hope I managed to preserve all the translation updates!
@jarrodmillman This should now be good to go.

steppi (approved):
Looks good. The rendered output for Portuguese and Japanese looks right to me just by comparing with what was there previously, @melissawm @AtsushiSakai.  The formatting changes might cause Crowdin to lose translations for the translated yaml files, but if so, I know how to fix that.

stefanv:
@steppi If you are happy with this PR, please go ahead and merge it. I'm hesitant, since it may impinge on your work.

stefanv:
Specifically, the structure of the `config.yaml` in https://github.com/numpy/numpy.org/pull/639 will be affected. But the update is also simple.

stefanv:
Thanks, let me know if you need a hand in fixing up the others!

steppi:
Thanks @stefanv. I'm pretty sure this won't cause problems

PR #653, opened 2023-07-21
Add Japanese translations

steppi:
This PR contains only the Japanese translations from gh-639. I used a scripted interactive rebase to filter commits by message. The script is in the details section below.

<details>

```bash

#!/usr/bin/env bash

# Check if a substring was provided
if [ "$#" -ne 1 ]; then
    echo "Usage: $0 substring"
    exit 1
fi

export substring=$1

# Save original values of GIT_SEQUENCE_EDITOR and GIT_EDITOR
orig_sequence_editor=$GIT_SEQUENCE_EDITOR
orig_editor=$GIT_EDITOR

# Set GIT_SEQUENCE_EDITOR to directly execute the grep and mv commands inline
export GIT_SEQUENCE_EDITOR="bash -c \"grep \\\"$substring\\\" \$1 > \$1.bak && mv \$1.bak \$1\""
export GIT_EDITOR=":"

# Perform the rebase
git rebase -i main

# Restore original values of GIT_SEQUENCE_EDITOR and GIT_EDITOR
export GIT_SEQUENCE_EDITOR=$orig_sequence_editor
export GIT_EDITOR=$orig_editor
```

</details>

@AtsushiSakai, please take a look to confirm everything is OK.

cc @rgommers @melissawm


AtsushiSakai:
@steppi Thank you for this preparation. Very useful.
I will check the Japanese translation in a few days based on the final rendering result.
Let me post some my findings about the final rendering here.

AtsushiSakai:
Should we translate the sample code comments to Japanese also?

image

AtsushiSakai:
Should we translate each tab name of the ecosystem?
image

AtsushiSakai:
Should we also translate the new letter sign up message?
image

steppi:
Thanks @AtsushiSakai for pointing out all of this text content which is missed by Crowdin. I don’t think it’s necessary to translate these things before we go live with the translations (hoping for next week) but it’s definitely worth considering for the near future. I’ll look into what we need to do to get Crowdin to pick these things up. 

For the newsletter sign up specifically, do you think translating that content could give the impression that the news letter is also available in translation? Though perhaps that’s something we also want to consider.

AtsushiSakai:
@steppi Thank you for your comment.
> we go live with the translations (hoping for next week)
Oh, great. I’m very happy to hear that. 

> I don’t think it’s necessary to translate these things
I totally agree that. I believe current translation is already helpful for Japanese people.

>do you think translating that content could give the impression that the news letter is also available in translation?
Ah, you're right. If this statement was translated, one might mistakenly think that the newsletter is also translated. When translating, it would be best to specify that the newsletter is in English.

I finished checking the whole rendering result, and found some untranslated sentence and some wrong translation.
I updated these in crowdin. Is it possible to merge the latest translation to this PR?
I hope next version might be a first release translation candidate.

steppi:
> I finished checking the whole rendering result, and found some untranslated sentence and some wrong translation.

Great, thanks for catching these @AtsushiSakai. 

> I updated these in crowdin. Is it possible to merge the latest translation to this PR?

Yes, that’s the plan. I’ll just cherry-pick the commits we need. I have another script for that. 

> I hope next version might be a first release translation candidate.

Sounds good. Let’s do it.

steppi:
@AtsushiSakai, the build failure was due to an inadvertently broken link. I've fixed it now. I'll merge after your approval.

AtsushiSakai (approved):
LGTM. I think there might be many small points to be improved, but I believe this is already useful resources for Japanese NumPy beginners.
@steppi Thank you for your kind support.

steppi:
> LGTM. I think there might be many small points to be improved, but I believe this is already useful resources for Japanese NumPy beginners. @steppi Thank you for your kind support.

Thanks @AtsushiSakai!

PR #652, opened 2023-07-18
Add an explicit anchor to releases section

steppi:
#### Brief description of what is fixed or changed
This PR adds an explicit anchor to the releases section of `news.md`. This is required for the link to work with translated content.

cc @rgommers @melissawm

PR #654, opened 2023-07-21
Add autotranslation of releases to build step

steppi:

Following up on @rgommers' comment https://github.com/numpy/numpy.org/pull/652#issuecomment-1640659743, this PR adds autotranslation of the list of releases to the build process. I'm marking this as draft because I think we should wait until after we go live with the first batch of translations before considering whether to merge this.

Autotranslation is accomplished with a simple Python script. I've briefly documented how to add new languages in a module level docstring for the script. The autotranslated releases can be included with a new shortcode `"layouts/shortcodes/releases.html"`.


rgommers:
Thanks Albert! This looks like the right approach to me from a quick look. Agreed with continuing after the translations launch. It may require a rebase or a few tweaks after that.

layouts/shortcodes/releases.html
```diff
@@ -0,0 +1,3 @@
+{{ $lang := .Page.Lang }}
```
stefanv, line 1: You may want to use one of the existing shortcodes provided by the theme:

https://theme.scientific-python.org/shortcodes/

steppi: Yeah, we could also just use the `include-md` short-code and in the `news.md` for each language, hard code the path to the `releases.md` for that particular language. Figuring out the language in the shortcode and generating the path dynamically might be too clever for its own good. Do you think the first option I mentioned with just `include-md` is better?

stefanv: Perhaps a simpler option is to generate a JSON data file, and to then use a language-specific template to generate the release lines:

https://gohugo.io/templates/data-templates/

steppi: OK, cool. Thanks @stefanv, I didn't know about this stuff! It looks like I've re-implemented something Hugo is already pretty good at. Data-templates does seem like the way to go.

steppi:
Closing this, because I plan to use @stefanv's data templates approach, and I think it will be easier to start from scratch.

PR #659, opened 2023-08-01
Update Hugo to 0.116.1

stefanv:


stefanv:
@steppi Looks like something's wrong with the Hugo 0.115.4 binary, but 0.116.1 seems to work OK.

steppi (approved):
OK. Looks good to me.

PR #661, opened 2023-08-01
Update GH pages workflow to not depend on external action

stefanv:
Use GitHub's new deploy action mechanism, getting rid of the action that relies on an Alpine Docker image, causing trouble with Hugo.


jarrodmillman (approved):

stefanv:
I'm going to merge this to see if it works. I am happy to revert if anyone is unhappy with it.

PR #663, opened 2023-08-01
Read Hugo version from netlify.toml

stefanv:


jarrodmillman (approved):

jarrodmillman:
You may want to remove this comment from `netlify.toml`:
```
# Same Hugo version as in .github/workflows/gh-pages.yml
```

PR #668, opened 2023-08-01
Update theme

jarrodmillman:

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,7 +2,7 @@
 # unless otherwise overridden by more specific contexts.
 [build.environment]
   PYTHON_VERSION = "3.8" # netlify currently only support 2.7 and 3.8
-  HUGO_VERSION = "0.116.1"
+  HUGO_VERSION = "0.118.2"
 
 # Here is another way to define context specific environment variables.
 [context.deploy-preview.environment]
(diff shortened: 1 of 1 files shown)


stefanv:
The alignment of the NumPy title doesn't look quite right to me.
I will take a look at what changed in the theme.

jarrodmillman:
It is getting close, but not quite there yet:
2023-08-18T12:28:49,997786884-07:00

alphapapa:
Thanks, investigating...

alphapapa:
With the change in https://github.com/alphapapa/scientific-python-hugo-theme/commit/30305d4d3d460548798fa5969f47f83d85497e7c, it now looks like this:
Screenshot_20230823_114800

What's the best way to effect this change here?  I suppose I should make a pull request on the scientific-python-hugo-theme repo to merge that change, and then after it's merged, this PR can be updated to point to that new commit, right?

jarrodmillman:
Thanks @alphapapa ! This is ready for review.

InessaPawson:
@alphapapa Thank you for working on this! 
@jarrodmillman @stefanv What do you  think of positioning the black button and the slogan above it further to the right, like in the screenshot below?

numpyorg-CTA-grid

Here is how it would look like without the grid:

numpyorg-CTA

alphapapa:
@InessaPawson Jarrod asked me to check on this.  It looks nice.  A few thoughts/questions:

1.  Would you share the CSS changes you made to do that?  
2. Does that change affect the positioning in narrow viewports, e.g. phone-sized, when the `div.flex-column` element is wrapped to be below the `div.hero-title` element?
3. I see that the `img.hero-logo` element is also smaller and appears like a superscript next to the `NumPy` text.  That looks nice as well, but if that were not also changed, it might affect the `div.flex-column` position (or its aesthetic appearance with regard to spacing).  Do you propose to make this change also?
4. Also, which browser did you take those screenshots in?

Thanks.

stefanv:
I think it would be good to ensure that (a) the logo appears as the smaller "superscript" version and (b) the space between the logo and button/text next to it remains (i.e., similar to how it currently looks on numpy.org).

Most sites seem to have good spacing between the two "columns", but because NumPy's box is so wide, it probably pushes into the logo. Options may be to wrap and stretch the box, or just to force a certain amount of spacing in between.

The logo changing size and position may have been an inadvertent theme change (or, at least, I didn't notice it until now).

InessaPawson (approved):

InessaPawson:
Thank you for working on this, @jarrodmillman ! Let’s merge it now and iterate, if needed, in followup PRs.

PR #670, opened 2023-08-02
Update dates in news entries

steppi:
This PR updates the dates for the latest news entries in news.md to the current date.


steppi:
@stefanv you can merge this one too if it looks good to you

stefanv:
Go ahead and self merge any fixes you need @steppi. We can always revert if there are any concerns.

PR #674, opened 2023-08-02
Fix overlap in text for Japanese translation

melissawm:

The Scientific Domain tab in the home page has overlapping text for the Japanese version (and also for the English version on mobile). These changes prevents overlap but are not responsive on mobile.

Before, on desktop:
Screenshot 2023-08-02 at 15 26 56

After:
Screenshot 2023-08-02 at 15 27 41

### DIFF
--- a/assets/css/tabs.css
+++ b/assets/css/tabs.css
@@ -109,6 +109,7 @@ tr.highlight-th {
 td.bold-text {
     font-weight: bold;
     text-align: center;
+    min-width: 120px;
 }
 
 .scientific-domains td.bold-text {


melissawm:
@stefanv maybe you folks have a better solution? I couldn't find where this width is defined on the CSS elements - I believe it is trying to determine the `td.text-center` cell width from the table column title width, so defining a minimal width for that should work. Unfortunately I am terrible at front-end stuff 😅

steppi:
I'm pretty sure the English site didn't have overlapping text on mobile prior to gh-673. Could "Processes" have been added as a hack to make the column widths right 😅

melissawm:
Rebased after #675

stefanv:
Thanks, @melissawm! This is a fine solution. In general, we should probably aim to move away from pixel sizing and towards `em` or `rem`, but that's a theme-wide concern.

PR #675, opened 2023-08-02
Update uikit

jarrodmillman:
Updated uikit to see if it might fix the issue discussed in #674. It doesn't seem to help, but it has a number of fixes compared to the older release and I don't see any new issues. It is only used for ecosystem tables.


melissawm (approved):
Certainly won't hurt 😄

jarrodmillman:
@melissawm It may make sense to merge this and rebase #674.

PR #683, opened 2023-09-19
update latest release button in all languages

mattip:
Fixes numpy/numpy#24744. Note the news fragment is out of sync: the English one highlights the 1.26 release where the Japanese and Portuguese still have an older news fragment.


seberg:
Thanks @mattip, LGTM.  Feels like it would be nice if this was clearer templated.

PR #686, opened 2023-10-25
Fix PyVista links

banesullivan:
Fixes some broken links to the PyVista project's documentation


InessaPawson (approved):

InessaPawson (commented):
Hi, @banesullivan! The existing link opens for me. However, if you prefer linking to PyVista's homepage, that will work too.

banesullivan:
The links might work for now (I think they are redirects), but I think for the long term it would be safer to link to the homepage

PR #703, opened 2023-12-03
Pin latest version of theme

stefanv:
@jarrodmillman Merging this so that the tabs on the front page work


jarrodmillman:
@stefanv Why do the tabs on the front page need the copy button?

Issue #707, opened 2023-12-07
Enable dark mode


jarrodmillman:
Here is a partial list of issues that need to be addressed:
  - [ ] Undo a85637404e6c80d808cd93271b8dbbe346f41761
  - [ ] Logos not rendering well in dark mode (scientific domains; sponsorships on About)
  - [ ] Tabs: last two tabs seem to have rendering problems; also check dark mode
`color`
  - [ ] Language drop down: wrong color in dark mode (should be fixed in theme?)

alphapapa:
Addressing in https://github.com/numpy/numpy.org/pull/708

autumn-absconds:
> Addressing in https://github.com/numpy/numpy.org/pull/708

Ok , but what might be the issue causing error in mine !?

alphapapa:
@jarrodmillman Do you still want to set the text color of the selected tab to match the unselected tabs, as it was before importing PDST, or do you want to keep it as PDST does it?

<details><summary>Screenshots</summary>
<p>

PDST style, as-is:  
![localhost_1313_](https://github.com/numpy/numpy.org/assets/601365/729b6962-480a-4a8f-8532-b5bc9e30408a)

Text matching other tabs: 
![localhost_1313_ (1)](https://github.com/numpy/numpy.org/assets/601365/f649af54-60dc-4070-bdac-a416ac7e067f)


</p>
</details>

stefanv:
Fine the way PDST does it.

autumn-absconds:
> @autumn-absconds This issue is for addressing the stated bugs.  What are you trying to do, and why?  i.e. are you trying to deploy the numpy.org Web site yourself, or...

I was trying to deploy myself to solve those bugs, read some guide earlier on how to start contributing to Open source , that's how it was to first clone and all that...

Till now I had used GitHub to store my small projects only ,
 This is my first time trying to do something like this , is this wrong way ?

alphapapa:
@autumn-absconds You are trying to learn about 20 things at once.  I would suggest that you slow down, pick a project that you already use personally, and see if there's some small issue you can work on.  This project has both volunteer and professional developers working on it, and we can't offer personal tutorials for new open source developers here on this issue report.  If you need more help, perhaps you could contact the author of the guide you read, or look for some IRC channel to help new developers.

autumn-absconds:
[...]

PR #708, opened 2023-12-07
Enable and fix for dark mode

alphapapa:
See https://github.com/numpy/numpy.org/issues/707.

- [x] Undo https://github.com/numpy/numpy.org/commit/a85637404e6c80d808cd93271b8dbbe346f41761: https://github.com/numpy/numpy.org/pull/708/commits/5dfb7969544515d7f8dd469865d9b53c12077ea5
- [x]  Logos not rendering well in dark mode (scientific domains; sponsorships on About):
  - https://github.com/numpy/numpy.org/pull/708/commits/bbf3708864445ad22a90b67a085c22ae42650aeb
  - https://github.com/numpy/numpy.org/pull/708/commits/b8a2621401c7982e58ffadc9bfb5cfdd6fe4ac25
  - https://github.com/numpy/numpy.org/pull/708/commits/ab211a573e0976de3739323666262007eb73778a
- [x] Tabs:
  * [x] "Array libraries" tab: logo images need filtering: 6c78d3bef7036bea0b1152d7dfafabec44935343
  * [X] "Data Science", "Machine Learning", "Visualization" tabs: colors need fixing: https://github.com/numpy/numpy.org/pull/708/commits/6c78d3bef7036bea0b1152d7dfafabec44935343
  * [x] "Visualization" tab's images: https://github.com/numpy/numpy.org/pull/708/commits/5b8a04b4d835045cfbaf86bb449e59d3f45eb54d
  * [x] Some tabs have a ~li~ bullet at the top of their contents: https://github.com/numpy/numpy.org/pull/708/commits/c0d23b7e08ab80b0588973793668b6ea6d7ecfa3
  * [x] Content top-margin larger for text than images: https://github.com/numpy/numpy.org/pull/708/commits/1d40832e5f7cdab405e67f6b2d63a461479169e6
- [x]  Language drop down: wrong color in dark mode (should be fixed in theme?)
  - [x] Fixed in https://github.com/scientific-python/scientific-python-hugo-theme/pull/435

### DIFF
--- a/assets/css/tabs.scss
+++ b/assets/css/tabs.scss
@@ -13,6 +13,10 @@ table td, table th {
 .tabs-section {
     display: flex;
     flex-direction: column;
+
+    html[data-theme=dark] & img {
+        filter: invert(1) !important;
+    }
 }
 
 .tabs-section .container {
@@ -133,10 +137,6 @@ img.first-column-layout {
     max-width: 100px;
     max-height: 30px;
     margin: 0px 20px 0px 10px;
-
-    html[data-theme=dark] & {
-        filter: invert(1) !important;
-    }
 }
 
 td.left-text {
--- a/assets/css/tabs.scss
+++ b/assets/css/tabs.scss
@@ -20,6 +20,14 @@ table td, table th {
 
     & .tabs div li {
         display: inline;
+
+        & .grid-container {
+            margin-top: 1em;
+
+            & p {
+                margin-top: 0em;
+            }
+        }
     }
 }
 
--- a/assets/css/tabs.scss
+++ b/assets/css/tabs.scss
@@ -133,6 +133,10 @@ img.first-column-layout {
     max-width: 100px;
     max-height: 30px;
     margin: 0px 20px 0px 10px;
+
+    html[data-theme=dark] & {
+        filter: invert(1) !important;
+    }
 }
 
 td.left-text {
--- a/assets/css/tabs.scss
+++ b/assets/css/tabs.scss
@@ -109,6 +109,10 @@ section.scientific-domains {
                 width: 50px;
                 height: 50px;
                 margin-bottom: 0.5em;
+
+                html[data-theme=dark] & {
+                    filter: invert(1) grayscale(1) !important;
+                }
             }
             & ul {
                 align-content: left;
--- a/assets/css/tabs.scss
+++ b/assets/css/tabs.scss
@@ -17,6 +17,10 @@ table td, table th {
     html[data-theme=dark] & img {
         filter: invert(1) !important;
     }
+
+    & .tabs div li {
+        display: inline;
+    }
 }
 
 .tabs-section .container {
--- a/config.yaml.in
+++ b/config.yaml.in
@@ -14,7 +14,6 @@ markup:
 disableLanguages: []
 
 params:
-  colorScheme: light
   author:
     name: "NumPy team"
   images:


alphapapa:
@stefanv @jarrodmillman Can't request reviews on this repo, so tagging you.  Thanks.

stefanv:
See https://deploy-preview-708--numpy-org.netlify.app/about/; what do we do with those logos?

alphapapa:
> See https://deploy-preview-708--numpy-org.netlify.app/about/; what do we do with those logos?

Apologies, I should have checked the other pages as well.

I think those logos can be decently handled by filters, like the smaller ones on the tab sections.  

Marking as a draft until ready for review again.  Thanks.

jarrodmillman:
See #712.

alphapapa:
Looking at the logos again, as [Stefan asked](https://github.com/numpy/numpy.org/pull/708#issuecomment-1846311231).

Two options come to mind:

1.  For ones with poor contrast, we could use a CSS filter to invert the colors.  That would likely fix the contrast, although it would, well, invert the colors.  :)
2. We could also adjust the background color, or try using a drop-shadow filter, something like that, to gently improve contrast without adjusting the image itself.

By the way, I noticed that the `sponsors.html` shortcode includes this stylesheet, which isn't functional because it's not in the page header:

https://github.com/numpy/numpy.org/blob/3c39fcfe02f94459e0456180b477ff6760f027d1/layouts/shortcodes/sponsors.html#L16-L43

The `partners.html` file includes one as well:

https://github.com/numpy/numpy.org/blob/3c39fcfe02f94459e0456180b477ff6760f027d1/layouts/shortcodes/partners.html#L13-L40

We should probably replace those with either inline styles on the relevant elements, or generalize them into the page's stylesheet.  

@stefanv @jarrodmillman 

Please let me know what you'd prefer that I do about them.  Thanks.

stefanv:
Sometimes, images can be used in light and dark mode, but we definitely need the ability to specify different images for each. How about something like this:

```html
<img src="img-dark.png" class="dark-theme"/>
<img src="img-light.png" class="light-theme"/>
```

Then, in the stylesheet:

```css
html[data-theme="dark"] {
  .light-theme {
    visibility: none;
  }
}

html[data-theme="light"] {
  .dark-theme {
    visibility: none;
  }
}
```

stefanv:
Re: `sponsors.html`, it should get its own `.css` file; or we remove `sponsors.html` and att it to `numpy.org`, since I think that's the only place it is being used.

alphapapa:
> Re: `sponsors.html`, it should get its own `.css` file; or we remove `sponsors.html` and att it to `numpy.org`, since I think that's the only place it is being used.

Sorry, I'm not sure what you mean about moving `sponsors.html` to `numpy.org`.

alphapapa:
> Sometimes, images can be used in light and dark mode, but we definitely need the ability to specify different images for each. How about something like this:

Sure.  I think we could also use the [picture element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) to do the same thing without having to add CSS classes.  What do you think?

alphapapa:
Pushed https://github.com/numpy/numpy.org/pull/725 in re https://github.com/numpy/numpy.org/pull/708#issuecomment-1936493405

jarrodmillman:
The style stuff in the shortcodes appears to be functional to me. It appears in the source file and when you remove the styling from the shortcode the page looks different. For example, with the styling
2024-02-09T14:17:22,330802550-08:00
and without it
2024-02-09T14:18:46,025201639-08:00

So we can probably just leave it as is for now.

alphapapa:
> The style stuff in the shortcodes appears to be functional to me. It appears in the source file and when you remove the styling from the shortcode the page looks different. For example, with the styling 2024-02-09T14:17:22,330802550-08:00 and without it 2024-02-09T14:18:46,025201639-08:00
> 
> So we can probably just leave it as is for now. Or move it.

Hm, ok, according to [this](https://stackoverflow.com/questions/4606708/declare-css-style-outside-the-head-element-of-an-html-page), it's not officially supported yet, but unofficially it works.

I'll close #725 then.  Thanks.

stefanv:
I considered that, but we do not merely rely on them preference, we also allow setting it, and that appears as a data element on html.

alphapapa:
> I considered that, but we do not merely rely on them preference, we also allow setting it, and that appears as a data element on html.

Thanks.

jarrodmillman:
Once we get the grid sorted out, maybe these can just be grids of images (i.e., maybe we don't need special roles for sponsors and partners).

jarrodmillman:
Let's put this on hold for a bit b/c I want to revisit if we can use theme functionality directly.

jarrodmillman:
See https://github.com/scientific-python/scientific-python-hugo-theme/pull/599

assets/css/tabs.scss
```diff
@@ -109,6 +125,10 @@ section.scientific-domains {
                 width: 50px;
                 height: 50px;
                 margin-bottom: 0.5em;
+
+                html[data-theme=dark] & {
```
shoumikchakravarty-dev, line 129: html[data-theme="dark"] is the recommended convention

assets/css/tabs.scss
```diff
@@ -13,6 +13,22 @@ table td, table th {
 .tabs-section {
     display: flex;
     flex-direction: column;
+
+    html[data-theme=dark] & img {
+        filter: invert(1) !important;
```
shoumikchakravarty-dev, line 18: invert(1) is a brute-force visual hack that doesn't scale. Try class-based escape by adding no-invert would let individual images opt out

assets/css/tabs.scss
```diff
@@ -13,6 +13,22 @@ table td, table th {
...
+    }
+
+    & .tabs div li {
```
shoumikchakravarty-dev, line 21: this looks to be unrelated to this PR

PR #713, opened 2023-12-20
Add a news article about the NumFOCUS EOY fundraiser

InessaPawson:


rgommers (approved):
Thanks @InessaPawson, this seems small enough and for a good cause - let's get this up. 

Could you revert the change to the address bar after the action expires on Dec 23rd?

InessaPawson:
@rgommers Will do.

PR #714, opened 2023-12-26
Update news bar to point to latest feature release again

InessaPawson:
Changing info in the news bar on the home page to the announcement about the 1.26 release since the NF EOY fundraiser ended on Dec 23, 2023.


rgommers (approved):
Thanks Inessa.

PR #728, opened 2024-02-20
Move keyfeatures to a grid of cards

jarrodmillman:
Building on functionality from the v0.11 release. The keyfeature panel for the homepage will be removed in a future theme release. This is part of our large effort to adopt styling from pydata-sphinx-theme where possible. It also makes the frontpage behave more like the other pages.

You can see more examples of cards (in a grid) here:
-  https://theme.scientific-python.org/shortcodes/#grid

Here is what they look like in PST:
- https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/web-components.html#cards

This will also make it easier to modify project homepages without having to override layouts/index.html.

The styling is a little different than before due to adopting pydata-sphinx-theme styling. But that styling is being used on many community sites and is being reviewed for accessibility.

The Scientific Python websites and scipy.org have already been updated. For example, see https://github.com/scipy/scipy.org/pull/558

@steppi @rgommers Note that this change will need to be made for all translations going forward.

(Once this is merged we plan to start converting the other special homepage sections---i.e., shell, casestudies, and tabs---to shortcodes called from content/<lang>/_index.md.)


InessaPawson (approved):
@jarrodmillman All the changes look great!

PR #735, opened 2024-03-15
Update theme (v0.15 and v0.14)

jarrodmillman:
@steppi @rgommers FYI, this updates the grid and figure shortcodes.

### DIFF
--- a/content/en/_index.md
+++ b/content/en/_index.md
@@ -2,7 +2,7 @@
 title:
 ---
 
-{{< grid1 columns="1 2 2 3" >}}
+{{< grid columns="1 2 2 3" >}}
 
 [[item]]
 type = 'card'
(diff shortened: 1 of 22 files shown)


jarrodmillman:
I used this script to convert between the format for Hugo's figure shortcode and the theme's figure shortcode:
```
import shlex
import sys


def replace_figure_shortcode(figure):
    figure = figure.replace('align="middle"', 'align="center"')
    contents = figure[11:-4]
    parameters = shlex.split(contents)
    parameters = [s for s in parameters if not s.startswith("class=")]
    toml = [s.replace("=", " = '") + "'" for s in parameters]
    return "\n".join(["{{< figure >}}"] + toml + ["{{< /figure >}}\n"])


filename = sys.argv[1]
start = "{{< figure src="
end = " >}}\n"


with open(filename, "r") as f:
    lines = f.readlines()

with open(filename, "w") as f:
    for line in lines:
        if line.startswith(start) and line.endswith(end):
            line = replace_figure_shortcode(line)
        f.write(line)
```

PR #736, opened 2024-03-16
Add HyperSpy

jlaehne:
Add the HyperSpy project to the ecosystem as group of libraries that draw heavily on `numpy` and several of its dependents such as `scipy`, `dask` and `sparse`
1. Under "Array libraries" as it provides tools for multi-dimensional data, that can be separated in signal and navigation dimensions
2. Under "Scientific Domains / Signal Processing" as the various HyperSpy extensions provide data analysis features for signals from a number of scientific domains, in particular from the realm of material sciences and electron microscopy

### DIFF
--- a/content/en/tabcontents.yaml
+++ b/content/en/tabcontents.yaml
@@ -29,7 +29,7 @@ params:
       alttext: JAX
       url: https://github.com/google/jax
     - title: Xarray
-      text: Labeled, indexed multi-dimensional arrays for advanced analytics and visualization
+      text: Labeled, indexed multi-dimensional arrays for advanced analytics and visualization.
       img: /images/content_images/arlib/xarray.png
       alttext: xarray
       url: https://xarray.pydata.org/en/stable/index.html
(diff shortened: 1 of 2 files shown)


rgommers (commented):
Thanks @jlaehne. There is one entry left under "Signal Processing" and HyperSpy seems like a nice fit there. 

I'd prefer to only have that entry; the second one under "Array Libraries" seems to not fit so well. HyperSpy is pretty specific to signal processing and microscopy/spectroscopy, basically adding 1D/2D objects to represent signals which wrap NumPy/CuPy as far as I can tell. It doesn't look like a generic array library.

jlaehne:
Thanks @rgommers for your review. Adding it also to "Array Libraries" was suggested by @ericpre, but I think it is good enough to have it listed under signal processing. I adapted the PR accordingly.

rgommers (approved):
Thanks @jlaehne, LGTM now - in it goes.

PR #744, opened 2024-04-12
Update tabcontents.yaml fix URL redirect http -> https

partev:
fix URL redirect http -> https

### DIFF
--- a/content/en/tabcontents.yaml
+++ b/content/en/tabcontents.yaml
@@ -84,7 +84,7 @@ params:
       alttext: A computer chip.
       img: /images/content_images/sc_dom_img/quantum_computing.svg
       links:
-        - url: http://qutip.org
+        - url: https://qutip.org
           label: QuTiP
         - url: https://pyquil-docs.rigetti.com/en/stable
           label: PyQuil
--- a/content/es/tabcontents.yaml
+++ b/content/es/tabcontents.yaml
@@ -79,7 +79,7 @@ params:
         alttext: Un chip para computador.
         img: /images/content_images/sc_dom_img/quantum_computing.svg
         links:
-          - url: http://qutip.org
+          - url: https://qutip.org
             label: QuTiP
           - url: https://pyquil-docs.rigetti.com/en/stable
             label: PyQuil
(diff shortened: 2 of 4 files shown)


content/en/tabcontents.yaml
```diff
@@ -89,7 +89,7 @@ params:
       alttext: A computer chip.
       img: /images/content_images/sc_dom_img/quantum_computing.svg
       links:
-        - url: http://qutip.org
+        - url: https://qutip.org
```
jarrodmillman, line 92: @rgommers @steppi It looks like this isn't in `content/pt/tabcontents.yaml` or `content/ja/tabcontents.yaml`. Is there a plan or mechanism to keep the translations up-to-date?

rgommers: @steppi could you please have a look at this?

stefanv: @steppi Gentle ping.

steppi: Sorry. I'm not sure how I missed the pings back in April. This url has made it into `content/ja/tabcontents.yaml` and `content/pt/tabcontents/yaml` now

https://github.com/numpy/numpy.org/blob/f9a497cb2046c2ae9c9bb992384577f1ca4ddaab/content/ja/tabcontents.yaml#L102-L103

https://github.com/numpy/numpy.org/blob/f9a497cb2046c2ae9c9bb992384577f1ca4ddaab/content/pt/tabcontents.yaml#L87


The mechanism for keeping translations up to date is that updates to the source content propagate to Crowdin through the GitHub integration and translators get notified that there is new or updated content to translate. The updates made in Crowdin are pushed to the running branch `l10n_main` and we have a GitHub action which essentially cherry picks commits from this branch to create PRs for particular languages.

Process-wise, it would be fine either to just update the url in `content/en/tabcontents.yaml` and let the other updates be made by translators through Crowdin, or to make the change for all of the translated urls directly in GitHub. Originally, we had Crowdin configured in such a way that made it difficult to resolve conflicts when changes are made outside of Crowdin but that isn't an issue any more. 

In this particular case, I think it's better the change `http` -> `https` directly in GitHub for the different languages, since this isn't really something that should require translator attention. As an aside, I want to figure out how to hide things like urls and other strings which should not be translated from the translators.

stefanv: Thank you, Albert!

stefanv (approved):

stefanv:
Thanks @partev!

PR #752, opened 2024-05-26
remove unnecessary single bullet point rendered under Ecosystem tabs

partev:
this removes unnecessary bullet point from "ECOSYSTEM" sections.
e.g. before the change:
image
and after:
image


rgommers (approved):
I'm not seeing that bullet on https://numpy.org/ in Firefox, but do see it on Chrome. So there is some browser dependent behavior there for some reason. After this PR, that seems to be solved indeed. So in it goes - thanks @partev!

PR #756, opened 2024-05-26
Update learn.md: use the latest edition of the "Python Data Science H…

partev:
…andbook"


rgommers (approved):
Thanks again @partev!

PR #757, opened 2024-05-26
this is a test do not merge yet

partev:
replace Twitter -> X


rkern:
I suggest giving your PRs meaningful titles about what they do. If they are not ready for review, instead of saying so in the title, just use the Github interface to mark it as a Draft PR.

partev:
> I suggest giving your PRs meaningful titles about what they do. If they are not ready for review, instead of saying so in the title, just use the Github interface to mark it as a Draft PR.

sure. I was trying to replace Twitter icon with X icon, but the build failed as I was afraid it would do. I will try to figure out how to do draft PRs.

Issue #498, opened 2021-11-13
Contribute page does not have link to donation


stefanv:
And this may be where people land looking for it.

InessaPawson:
Fair point! We should definitely add a paragraph about it and links/buttons.

blackhat06:
I can own this issue and contribute to it

melissawm:
Hi @blackhat06 - feel free to try! I think there are still a couple of things we should decide, though.

1. We should add a link to https://numpy.org/about/#donate in the https://numpy.org/contribute page
2. The link to donate through NumFOCUS in https://numpy.org/about/#donate points to [this page](https://numfocus.org/donate-to-numpy), which looks outdated and should probably be replaced by https://opencollective.com/numpy
3. The links in https://numpy.org/about/#donate should probably reflect the same content as the `Sponsor this project` section on https://github.com/numpy/numpy

Does this seem reasonable?

blackhat06:
These seems reasonable to me.

BhaswatiRoy:
Hello @melissawm 
I can work on this issue

PR #768, opened 2024-08-11
Add info about donating to NumPy to contribute.md

InessaPawson:
Add info about donating to NumPy to contribute.md
Resolves https://github.com/numpy/numpy.org/issues/498

Improve Fundraising section

### DIFF
--- a/content/en/contribute.md
+++ b/content/en/contribute.md
@@ -107,18 +107,6 @@ Fundraising is a scarce skill here -- we'd appreciate your help.
 
 ### Donate
 
-If you'd like to contribute to NumPy by making a donation, you are welcome to do so. 
-Any amount helps! NumPy Steering Council will make the decisions on how to best use any funds received. 
-Technical and infrastructure priorities are documented on 
-the [NumPy Roadmap](https://www.numpy.org/neps/index.html#roadmap).
+If you'd like to contribute to NumPy by making a donation, visit [https://numpy.org/about/#donate](https://numpy.org/about/#donate).
 
-NumPy is a Sponsored Project of NumFOCUS, a 501(c)(3) nonprofit charity in the United 
-States. NumFOCUS provides NumPy with fiscal, legal, and administrative support to help 
-ensure the health and sustainability of the project. 
-Visit [numfocus.org](https://numfocus.org) for more information.
 
-Donations to NumPy are managed by [NumFOCUS](https://numfocus.org). For donors in 
-the United States, your gift is tax-deductible to the extent provided by law. As 
-with any donation, you should consult with your tax advisor about your particular tax situation.
-
-{{<opencollective>}}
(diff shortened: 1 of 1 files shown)


stefanv:
I wonder if we should encourage small donations. Ultimately, recurring larger contributions (eg Tidelift), grants, etc. allow us to hire contractors. But I feel bad that we should fund-raise small amounts from people who don't have a lot. Better perhaps for them to help out in other ways?

content/en/contribute.md
```diff
@@ -105,10 +98,27 @@ sprints](https://scisprints.github.io/), a newsletter, and perhaps a blog.
...
+For many years, NumPy was maintained by dedicated volunteers, but as its importance grew it 
+became clear that to ensure stability and growth we would need financial support.
+[This SciPy'19 talk](https://www.youtube.com/watch?v=dBTJD_FDVjU) explains how much difference 
+that support has made. Like most nonprofits, we are constantly seeking grants, sponsorships, 
+and other kinds of funding. We have a number of ideas and of course we welcome more. 
+Fundraising is a scarce skill here -- we'd appreciate your help.
+
+### Donate
+
+If you'd like to contribute to NumPy by making a donation, you are welcome to do so.
```
rgommers, line 110: It would be better to have one sentence here linking to `about.md#donate` (see https://numpy.org/about/#donate), rather than duplicate this content.

rgommers: That content on the About page also needs a small update, to point to Open Collective (as pointed out in https://github.com/numpy/numpy.org/issues/498#issuecomment-1295930356)

rgommers (commented):
Thanks Inessa!

> I wonder if we should encourage small donations. Ultimately, recurring larger contributions (eg Tidelift), grants, etc. allow us to hire contractors. But I feel bad that we should fund-raise small amounts from people who don't have a lot. Better perhaps for them to help out in other ways?

I can see arguments for this either way. However, this PR is more a minor maintenance update than a significant change, since we already have this content on the About page.

On the one hand, at the moment we don't have needs that we cannot cover with the ~$80,000 project balance (see https://opencollective.com/numpy). The one source of recurring unrestricted income for the project is $1,000/month from Tidelift (and that amount has been declining over time).

On the other hand, I can imagine that the good times (multiple maintainers getting significant day job time from their employers, as well as always >=1 active grant since 2017) won't last forever. And grant funding tends to not be unrestricted (luckily quite a bit of employer day job time is).

* * *

I'd merge this PR after comments are addressed, because it doesn't add anything that wasn't already there.

content/en/contribute.md
InessaPawson: @rgommers Thank you for your feedback! The button in `about.md#donate` is currently pointing to https://opencollective.com/numpy#category-CONTRIBUTE Did you have a different "destination" in mind?

rgommers: Sorry for the noise, the link is fine. I was either looking at an old page or had some temporary blindness - that comment was addressed a long time ago. Please ignore this part.

rgommers (approved):
LGTM now, in it goes. Thanks Inessa! And thanks for the review Stefan.

PR #772, opened 2024-08-31
Add gh workflow for creating translations PR

steppi:

This PR adds a GitHub workflow for creating a pull request to add or update translations for a given language. It has a `workflow_dispatch` which accepts a [Crowdin language code](https://developer.crowdin.com/language-codes/). The workflow clones the `numpy.org` repo, ensures that there are no merge conflicts between the translation branch `l10n_main` and `main`, checks out `l10n_main` and then does a scripted interactive rebase onto `main` which picks only commits for the language of interest. It uses scripts from the repo https://github.com/Scientific-Python-Translations/automations, primarily [create_branch_for_language.sh](https://github.com/Scientific-Python-Translations/automations/blob/main/scripts/create_branch_for_language.sh) to accomplish it's work.

I've been testing this out on my fork, and have verified that it works there. See https://github.com/steppi/numpy.org/actions/runs/10646298865/job/29512984917, and https://github.com/steppi/numpy.org/pull/6 for an instance of a PR generated by this workflow.

The goal here is to make it easier to create pull requests to add translations, moving something that was a bespoke thing I did on my local machine, to a push button thing anyone with the right permissions can do.

For this to work as is, the `GITHUB_TOKEN` needs to be given write permissions and permission to create pull requests in the actions settings, if it doesn't already have these permissions.

image

cc @melissawm


steppi:
Would it be better to make the PR from a branch on a `scientific-python-translations/numpy.org` fork, rather than a branch on`numpy/numpy.org` directly?

melissawm (commented):
Sorry for the delay @steppi - this looks fine to me but I'd love to see an example of a crowdin commit so I can wrap my head around the language codes and branches. Do you have that handy?

melissawm:
> Would it be better for the workflow to make the PR from a branch on a `scientific-python-translations/numpy.org` fork, rather than a branch on`numpy/numpy.org` directly?

My gut feeling is that this would increase the chances of something getting out of sync, no?

steppi:
Thanks @melissawm for taking a look! 

I'm going to go ahead and merge this. I guess a new workflow which can only be triggered manually through a workflow dispatch by a maintainer can't be too harmful. At worst, I imagine it just might not work due to some difference in the configuration here verse on my fork.

PR #784, opened 2024-09-18
Fix display of ecosystem tab for Ja and Pt

melissawm:

This restores the display of the Scientific Domains tab for Japanese and Portuguese. An update of how these tabs are pulled together was not applied to the translated pages which was causing the missing content.

Also updates the CSS for the tabs so that the scientific domain titles are correctly aligned.

Finally, there were some newlines applies to yaml files that are probably incorrect.

### DIFF
--- a/assets/css/tabs.scss
+++ b/assets/css/tabs.scss
@@ -95,6 +95,7 @@ section.scientific-domains {
             line-height: 130%;
             margin: 0.2em 0.4em;
             flex-basis: 13%;
+            align-self: baseline;
 
             & header {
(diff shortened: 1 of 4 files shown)


melissawm:
Seems to work https://deploy-preview-784--numpy-org.netlify.app/pt/

steppi (approved):
Thanks @melissawm. Looks good

stefanv:
Go ahead and merge when you're ready, @melissawm

steppi (approved):

PR #797, opened 2024-12-08
audit workflows via zizmor and correct use of env variables

mattip:
Recently there was a [hack of a repo](https://blog.yossarian.net/2024/12/06/zizmor-ultralytics-injection) that used environment variables unsafely. While this repo "only" publishes documentation, it is better to tighten things up where possible. There is a tool called [zizmor](https://woodruffw.github.io/zizmor/) that can be used to audit workflows, I used it to develop this PR:
- use [persist-credentials: false]  in the checkout step, see https://github.com/actions/checkout/issues/485
- use quoted environment variables instead of directly using user-provided values in scripts


rgommers (approved):
This all LGTM, in it goes. `zizmor` looks like a potentially useful tool. Thanks @mattip

PR #806, opened 2024-12-17
Add translations team

### LINKED ISSUE #796: Add a "Translation Team" to the teams page
I was thinking about the best way to credit translators for the NumPy.org page, and I thought this might be a good idea. I'd like to propose using the Teams gallery page provided by the scientific python hugo theme to show a "Translation Team".

The only issue is that right now the Teams page is sourced directly from GitHub teams, and there are two problems there:
- There is no translation team on github (but this would be easily fixed)
- Not every translator has a github account - they log in to Crowdin with potentially other credentials.

If there is interest and consensus, could we have a manual section on the teams page to add translators? Or source them directly from crowdin?

cc @steppi

melissawm:

Hi all! This PR adds a "Translations team" section to the teams page. The toml file is manually added since translators are members of crowdin but not necessarily github.

I am happy to add links to github profiles and/or new avatars, just let me know. As a placeholder, folks without an avatar are showing up as the NumPy logo.

Closes #796

### DIFF
--- a/content/en/teams/index.md
+++ b/content/en/teams/index.md
@@ -27,6 +27,10 @@ communities worldwide by building quality, open-source software.
 
 {{< grid file="survey-team.toml" columns="2 3 4 5" />}}
 
+### Translations team
+
+{{< grid file="translations-team.toml" columns="2 3 4 5" />}}
+
 ### Emeritus maintainers
 
 {{< grid file="emeritus-maintainers.toml" columns="2 3 4 5" />}}
--- a/content/en/teams/translations-team.toml
+++ b/content/en/teams/translations-team.toml
@@ -15,8 +15,8 @@ body = '''{{< image >}}
 src = 'https://raw.githubusercontent.com/numpy/numpy.org/refs/heads/main/static/images/logo.svg'
 alt = 'Avatar of Juan Pablo Duque'
 {{< /image >}}
-Juan Pablo Duque'''
-link = '@juanpabloduqueo'
+Juan Pablo Duque (@juanpabloduqueo)'''
+link = 'https://scientific-python.crowdin.com'
 
 [[item]]
 type = 'card'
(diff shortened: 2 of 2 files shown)


stefanv:
Hi @melissawm, is it possible to make the crowdin project public so we can automatically scrape team members? AFAI understand, you can make a project public without automatically allowing volunteers to join; but I don't know if we can see team members of a public project; worth looking, though?

rgommers:
I'm not sure about making the Crowdin project public - perhaps there is value in that in itself - however automatically scraping team members doesn't seem desirable. Translators get added immediately on request, but we only want to credit them as part of the NumPy Translations Team after a substantial set of contributions, just like there is a bar for being listed as part of other teams.

rgommers:
> I am happy to add links to github profiles and/or new avatars, just let me know. As a placeholder, folks without an avatar are showing up as the NumPy logo.

The links should go to GitHub profiles (when available) in my opinion, and then the profile images will show up so this will start looking more normal:

image

rgommers:
Thanks for doing this Melissa, +1 for crediting translators!

melissawm:
@stefanv we can definitely investigate that, but depending on the effort involved in scraping maybe that could be a follow-up? I think having the translators credited helps with both engagement for new translators and actual rewards for the work done, so I would not like to let it sit for too long 😄

stefanv:
OK, I'll merge if you'll investigate that option 🤝 🙂

stefanv:
Sorry, I missed Ralf's comment:

> but we only want to credit them as part of the NumPy Translations Team after a substantial set of contributions, just like there is a bar for being listed as part of other teams.

In that case, fine to keep this as-is.

stefanv:
(BTW profile currently links to the crowdin page, which most people can't see.)

melissawm:
@steppi do you know how to make the crowdin pages for the scientific python projects be public?

PR #815, opened 2025-01-05
Fix date format in news.md

Musharraffaijaz:

### DIFF
--- a/content/en/news.md
+++ b/content/en/news.md
@@ -1,8 +1,8 @@
 ---
-title: "News Title"
+title: "News"
 sidebar: false
 newsHeader: "NumPy 2.2.0 released!"
-date: 2023-10-15
+date: 2024-12-08
 ---
 
 ### NumPy 2.2.0 released
--- a/content/en/news.md
+++ b/content/en/news.md
@@ -1,8 +1,8 @@
 ---
-title: News
+title: "News Title"
 sidebar: false
 newsHeader: "NumPy 2.2.0 released!"
-date: 2024-12-8
+date: 2023-10-15
 ---
 
 ### NumPy 2.2.0 released


melissawm (commented):
Hi @Musharraffaijaz - I don't understand these changes. We want the title of the page to be just "News" so maybe adding the quotes is what is needed. As for the date, maybe we could write `2024-12-08` but we should not change the actual date of the published news.

I'm curious to understand why this PR is needed as the page seems to be rendering correctly: https://numpy.org/news/

Musharraffaijaz:
> Hi @Musharraffaijaz - I don't understand these changes. We want the title of the page to be just "News" so maybe adding the quotes is what is needed. As for the date, maybe we could write `2024-12-08` but we should not change the actual date of the published news.
> 
> I'm curious to understand why this PR is needed as the page seems to be rendering correctly: https://numpy.org/news/

Hi @melissawm hope you are doing well. Sorry for the inconvenience made, but without the made changes it showed me the following error:
Screenshot From 2025-01-06 13-21-15

Quoting the error: 
>ERROR the "date" front matter field is not a parsable date: see /home/vboxuser/Documents/SoD/numpy.org/content/en/news.md

So even after trying your suggestions of just adding the quotes and changing it to "News" didn't made any difference to the error. The issue is with the date, as in the "2024-12-8"  it is missing a leading zero in the day part, and that's why it was a trail and error from my end, which was successful in the first attempt. 
After changing the date to **_2024-12-08_** resolved the issue, Again sorry for the inconvenience made

melissawm:
No worries - if you change your PR to do just that (add the 0 to the date) we could merge. Thanks!

melissawm:
Thanks!

Musharraffaijaz:
Can you kindly review the [#816](https://github.com/numpy/numpy.org/pull/816) as well. This PR is related to the update in readme.md file.

PR #819, opened 2025-01-16
Remove dash from main page title

### LINKED ISSUE #766: remove an extra dash from numpy.org website's title "NumPy - "
current Numpy.org's website title is "NumPy - "

It is not clear what the purpose of the extra dash is. It feels like something else, like a short description of NumPy was meant to follow after the dash but is somehow missing. I suggest either removing the dash, or adding a short description after the dash.

Screenshot 2024-07-22 01 08 29

jarrodmillman:
Close #766

### DIFF
--- a/content/en/_index.md
+++ b/content/en/_index.md
@@ -1,5 +1,5 @@
 ---
-title:
+title: NumPy
 ---
 
 {{< grid columns="1 2 2 3" >}}
--- a/content/es/_index.md
+++ b/content/es/_index.md
@@ -1,5 +1,5 @@
 ---
-title: null
+title: NumPy
 ---
 
 {{< grid columns="1 2 2 3" >}}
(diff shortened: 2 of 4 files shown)


stefanv:
@steppi I'd like to merge this, but I noticed in another PR you noted that we should not be editing translated files. Could you clarify whether it is OK for us to edit translations here, to correct technical issues like this one (and, before, like fixing the format of image captions, e.g.)?

steppi:
> @steppi I'd like to merge this, but I noticed in another PR you noted that we should not be editing translated files. Could you clarify whether it is OK for us to edit translations here, to correct technical issues like this one (and, before, like fixing the format of image captions, e.g.)?

Thanks for the ping @stefanv. This is fine to merge. I know how to clean up the conflicts in Crowdin now. Touching translations should be fine, afterwards we just have to delete the `110n_main` branch that Crowdin creates. It will then attempt to create a new one with conflicts resolved. Crowdin is configured in such a way that translations won't be lost during this process, although this wasn't the case previously.

PR #822, opened 2025-01-19
Announce NumPy 2.2.2

charris:


stefanv:
Failure is coming from previously merged translations which erroneously altered links /cc @steppi

stefanv:
Thanks for the release, Chuck!

PR #824, opened 2025-01-21
Disable code execution on page load for the NumPy REPL

agriyakhetarpal:

This pull request disables the code execution for the `import numpy as np` command, which should reduce bandwidth usage as NumPy would then not be downloaded from the Pyodide CDN as a result until the user decides to press the <kbd>Enter</kbd>or <kbd>return</kbd> buttons (or any other buttons to inject code into the REPL).

This useful option comes from https://github.com/jupyterlite/jupyterlite/pull/1547, which was released in https://github.com/jupyterlite/jupyterlite/releases/tag/v0.5.0, and was updated to in https://github.com/jupyterlite/demo/pull/150, i.e., with the JupyterLite deployment that we use.

<hr>

cc: @rgommers


stefanv (approved):
Thanks @agriyakhetarpal; that's a good way to save some resources.

stefanv:
See https://github.com/numpy/numpy.org/pull/824 for a CI fix.

stefanv:
Thank you @agriyakhetarpal!

PR #825, opened 2025-01-21
Use plain ascii anchors to avoid problems with link checker

stefanv:
@steppi I think the translated anchors were correct, but suspect the netlify link checker does not deal perfectly with unicode. I made the anchors plain text ascii to see if we can work around the issue.


stefanv:
We may want to consider explicitly setting all IDs to the English anchor, so that you can navigate between languages by just changing `en` to `pt` etc. But, I'll leave it up to you to decide if that is desirable.

jarrodmillman (approved):
Thanks!

Issue #811, opened 2024-12-23
DOC: update installation page


lucascolley:
We just overhauled the SciPy installation page at https://scipy.org/install/, including:
- creating a separate beginner guide for those who may not be comfortable with using a terminal or happy to learn to use a package manager
- recommending project-based workflows to users who do not consider themselves beginners (`uv` and `pixi`)
- modernising instructions for installing with `pip` or `conda`

You may wish to do similar things for https://numpy.org/install/. Feel free to copy anything I wrote on the SciPy page!

mattip:
It might be nice to set this up as a tab-based page with a tab for each installation method. This would emphasize that they are equivalent alternatives.

Musharraffaijaz:
I tried working on the same issues, but if you could kindly review the PR, Fixed the minor errors which hindered the development server #814 after which I have faced issues with the index page not avaliable, and shows **_Page not found_** 
![image](https://github.com/user-attachments/assets/a173830a-ca0a-4f87-8e72-36f871947f6d)

Musharraffaijaz:
Hi, 
@lucascolley @mattip I'm working on a PR that would involve making some lengthier commits to the installation instructions in the `install.md` file. 
If thats fine, I'd be happy to create the PR for your review.

Thanks,

Musharraffaijaz:
@mattip what about this? Since in markdowns, there is no option for Tabs and we can get tabs in MDX instead, so instead of a tab based navigation, how about the tabular one?  

![Image](https://github.com/user-attachments/assets/8c28a5e4-634c-4fbb-80dc-49533f238671)

mattip:
If you want you could try something like [this](https://yairgadelov.me/simple-tab-widget-for-hugo-blog/) for tabs. I think the tabular (table) view wastes too much valuable space.

Musharraffaijaz:
The PR  **[updated the installation page with tab based navigation #608](https://github.com/scipy/scipy.org/pull/608)**  resolves this issues

lucascolley:
Hi @Musharraffaijaz - it looks like your PR affects scipy.org. This issue was about potential updates to numpy.org.

I'll have a look at your PR to see whether it looks like an improvement for SciPy! But I don't think it would close this issue.

jarrodmillman:
Here is how you use tabs with the theme: https://theme.scientific-python.org/user_guide/web-components/#tabs

jarrodmillman:
[...]

PR #832, opened 2025-02-08
update the numpy installation page with tab based navigation 

Musharraffaijaz:
image

### DIFF
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -2,206 +2,126 @@
 title: Installing NumPy
 sidebar: false
 ---
-
+{{< admonition tip >}}
+This page assumes you are comfortable using a terminal and are familiar with package managers. 
 The only prerequisite for installing NumPy is Python itself. If you don't have
 Python yet and want the simplest way to get started, we recommend you use the
 [Anaconda Distribution](https://www.anaconda.com/download) - it includes
 Python, NumPy, and many other commonly used packages for scientific computing
 and data science.
+{{< /admonition >}}
 
-NumPy can be installed with `conda`, with `pip`, with a package manager on
+The recommended method of installing NumPy depends on your preferred workflow. Below, we break down the installation methods into the following categories:
+
+- **Project-based** (e.g., uv, pixi) *(recommended for new users)*
+- **Environment-based** (e.g., pip, conda) *(the traditional workflow)*
+- **System package managers** *(not recommended for most users)*
+- **Building from source** *(for advanced users and development purposes)*
+
+Choose the method that best suits your needs. If you're unsure, start with the **Environment-based** method using `conda` or `pip`.
+
+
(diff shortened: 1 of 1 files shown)


Musharraffaijaz:
This PR tries to resolve the issue **_[DOC: update installation page #811](https://github.com/numpy/numpy.org/issues/811#top)_**

mattip:
When I look at [the installation page from this PR](https://deploy-preview-832--numpy-org.netlify.app/install/) I get confused. The page opens with a section on conda and pip, and then there are three tabs with the titles "Recommended method", "Python Package Management", and "NumPy Packages and Libraries". The meaing of those titles is not clear, and the pip/conda directions are repeated. 

The layout on the [Scipy page](https://scipy.org/install/) is much better. Could you copy that page to here and replace `scipy` with `numpy`?

Musharraffaijaz:
> When I look at [the installation page from this PR](https://deploy-preview-832--numpy-org.netlify.app/install/) I get confused. The page opens with a section on conda and pip, and then there are three tabs with the titles "Recommended method", "Python Package Management", and "NumPy Packages and Libraries". The meaing of those titles is not clear, and the pip/conda directions are repeated.
> 
> The layout on the [Scipy page](https://scipy.org/install/) is much better. Could you copy that page to here and replace `scipy` with `numpy`?

Thank you for your review and reply, I have a small question, what about the installation steps? Since they are all different? So do you just want the scipy initial para to be the similar?
image

melissawm:
The NumPy installation should be simpler, but yes the page needs a full rewrite I'm afraid. You should verify that those steps work for NumPy as well, and there are a few that are only necessary for SciPy.

Musharraffaijaz:
@melissawm @mattip thank you for your suggestions, could you please check the latest commits? I have tried something different have checked and verified different installation methods accros internet and through GPT as well. Could you please review this and suggest changes. Thanks

content/en/install.md
```diff
@@ -2,195 +2,126 @@
...
-**CONDA**
+Choose the method that best suits your needs. If you're unsure, start with the **Environment-based** method using `conda` or `pip`.
 
-If you use `conda`, you can install NumPy from the `defaults` or `conda-forge`
-channels:
 
+NumPy can be installed with a `package manager` on
+macOS and Linux, or [from source](https://numpy.org/devdocs/building).
+For more detailed instructions, consult our **Python and NumPy
+installation** guide below.
```
mattip, line 27: I think these lines can be removed.

```suggestion
```

mattip (commented):
This looks really good, thanks. Just a small nit.

content/en/install.md
Musharraffaijaz: Thank you for reviewing it, have updated the PR with the said changes.

mattip:
Thanks @Musharraffaijaz

PR #795, opened 2024-11-18
Add link and image of NumPy comics to Contribute page

MarsBarLee:
I've added a link and image to the 'How to Contribute to NumPy' comic in the Contribute page.

However, the image is showing up as broken like this:
Screenshot 2024-11-18 at 2 23 50 PM

@Mukulikaa Could you take a look at this?

PR #834, opened 2025-02-10
Add shortcode to add a resized image of the NumPy comic

MarsBarLee:
Hi all,
 #795 already adds a link to the NumPy comic in the Contribute page.

I want to add a version that add an image of the comic. While I could add the image directly with #795, there is no way to easily resize the image across a variety of browser sizes and devices.

This shortcode adds a comic.html so that CSS can automatically resize the image. This format is used elsewhere in the NumPy

#### Alternative implementation

There is also an option to embed a flippable version of the comic directly on the Contribute page. This is how Heyzine (the flipbook host) provides their embed option through the `iframe` tag.
```html
<iframe allowfullscreen="allowfullscreen" scrolling="no" class="fp-iframe" style="border: 1px solid lightgray; width: 100%; height: 400px;" src="<https://heyzine.com/flip-book/3e66a13901.html>"></iframe>
``` 

I haven't tested out yet if NumPy website works nice with iframe.


Mukulikaa:
Thanks, @MarsBarLee. This looks good! Just a note: Mars will also try to embed it in the docs in a future PR.

PR #856, opened 2025-04-28
Add translations for Japanese

scientificpythontranslations:
This PR adds the translations for Japanese to the project.

This PR was automatically created by the @scientificpythontranslations bot and only commits that resolve any merge conflicts should be pushed directly to this branch/PR. Any modifications of the translated content should be addressed directly on the [Crowdin Project Site](https://scientific-python.crowdin.com//u/projects/12/l/ja).

The Crowdin integration for this repository is located at https://github.com/Scientific-Python-Translations/numpy.org-translations.


melissawm:
Hi @agriyakhetarpal - am I wrong that maybe https://github.com/scientific-python/scientific-python-hugo-theme/pull/700/changes broke the build here? It seems like the theme is looking for a discourse-comments anchor on the news page, but we don't use comments here. Can we disable this feature for NumPy? Thanks!

melissawm:
Ok I think I got it - I don't think we need a deep review of content; for the conflicts, I checked with some automatic translation just to make sure the strings were in the correct place. I would like to get this merged sooner rather than later so we can make sure the contents are finally synced.

cc @InessaPawson would you mind helping me here? Thanks ❤️

rgommers (approved):
Thanks @melissawm and all translators! This looks good - I reviewed the diff, and then reviewed the translations quickly by verifying in Chrome that they seemed reasonable when auto-translating in reverse (that's not as good as a native speaker review of course, but close enough to ensure that no accidents happened during the integration).

PR #864, opened 2025-05-23
Use current quansight logo

ngoldbaum:
The Quansight marketing team asked me to update the logo on the numpy website.


ngoldbaum:
current website:

image


proposed change:

image


tbh it looks a little less nice than it used to because the crop on the three logos isn't quite the same. Maybe there's a CSS tweak that would look nicer?

melissawm:
I was looking at this locally and could not find a css rule that solves all of it - basically because the Quansight logo has some whitespace at the top while the other logos don't:

Captura de imagem_20250523_115959

I think we can either crop the Quansight logo or add some padding to the other ones if we want this to look nice. I'm happy to look into it if you need help with the CSS stuff.

InessaPawson (commented):
Since this is in my wheelhouse, I’ll chime in. If the goal is maintainable code and consistent UI, I recommend removing the excess whitespace from the logo image.

ngoldbaum:
I cropped the whitespace a bit and now it looks better.

melissawm (approved):
Looks great to me, thanks @InessaPawson and @ngoldbaum !

PR #872, opened 2025-06-23
Change prompt cell position for the NumPy REPL (do not merge)

### LINKED ISSUE #869: DOC: Change the position of the prompt cell for the JupyterLite REPL?
### Issue with current documentation:

Currently the embeded REPL on the https://numpy.org landing page shows a REPL with the prompt cell at the bottom, which can be confusing to some users:

Image

### Idea or request for content:

As a quick heads-up, JupyterLite 0.6.0 (currently in beta) will allow changing the position of the prompt cell: https://jupyterlite.readthedocs.io/en/latest/changelog.html#enhanced-repl-options

Image

Just wanted to bring this up in case folks would be interested in tweaking a couple of settings for the current REPL.

cc @agriyakhetarpal who opened a similar issue on another repo (also in case a common single deployment for scientific Python is still something to consider).

agriyakhetarpal:

This PR changes the prompt cell position for the NumPy REPL from the default (`bottom`) to `left`. This is, at the moment, not intended to be merged, as it is solely for trying out the new feature that was introduced the recent JupyterLite 0.6 release. Here is a visual diff:

Scenario | Before | After |
|--:|:----:|:----:|
| With the default `import numpy as np` snippet | numpy org_ | numpy org_ (1) |
| The example snippet is executed | numpy org_ (2) | numpy org_ (3) |

This might currently remain blocked until https://github.com/jupyterlab/jupyterlab/issues/17597 is resolved (I previously attempted to fix this in JupyterLite, but I was later prompted to the fact that the fix lies in JupyterLab instead). I would be on the fence here personally, as both of the layouts look fine to me. However, if the line length in a code snippet is long, say, >70 characters, some of them might get hidden until the layout is expanded manually.

A follow-up item could be to introduce a `width`/`ratio` element (for the `left` or the `right` positions) which could control the initial state of the prompt cell panel's and the output side panel's widths – it's currently split 50-50 between both but I could see a use case for a 40-60 or 30-70 split (or vice versa).

cc: @jtpio

Closes gh-869

### DIFF
--- a/layouts/partials/shell.html
+++ b/layouts/partials/shell.html
@@ -17,7 +17,7 @@
             <!-- Interactive Shell -->
             <iframe
                 class="numpy-shell"
-                src="https://jupyterlite.github.io/demo/repl/?toolbar=1&kernel=python&execute=0&code=import%20numpy%20as%20np"
+                src="https://jupyterlite.github.io/demo/repl/?toolbar=1&kernel=python&promptCellPosition=left&execute=0&code=import%20numpy%20as%20np"
             >
             </iframe>
         </div>


InessaPawson:
@agriyakhetarpal Hi, Agriya! What is the status of this PR?

PR #882, opened 2025-10-31
Add a reference to Blosc2 as an array library

FrancescAlted:

This PR adds a reference to [Blosc2](https://www.blosc.org/python-blosc2/), a nascent array library.  Closes https://github.com/numpy/numpy/issues/30107.  I have added translations for Spanish (native speaker here), Portuguese (close enough to Spanish), but I did not attempt to provide the translation to Japanise to avoid japanese folks being offended.


FrancescAlted:
Thanks @rgommers !

PR #886, opened 2025-11-30
Fix improper comic shortcode that causes the NumPy navbar logo to be shrunk on the "Contribute" page(s)

agriyakhetarpal:

The comic shortcode currently applies a style to all images it is utilised on, so the "Contribute" pages also inherit a `max-width: 20%` rule, which downsizes the NumPy logo in the navbar. Additionally, the div that applies the `comic` class to the comic cover image was being foreclosed without including the comic cover image within itself, breaking the secondary sidebar as the HTML became syntactically invalid (for context, the secondary sidebar is visible on other pages, such as the "Community" and "About Us" pages). This PR corrects these issues. See the table below for a comparison.

| Before | After |
|:------:|:------:|
| image | image |

### DIFF
--- a/layouts/shortcodes/comic.html
+++ b/layouts/shortcodes/comic.html
@@ -7,6 +7,5 @@
 <style>
     .comic img {
         max-width: 20%;
-        max-height: 20%;
     }
 </style>


layouts/shortcodes/comic.html
```diff
@@ -1,16 +1,11 @@
...
     <a href="https://heyzine.com/flip-book/3e66a13901.html">
         <img src="/images/content_images/numpy-comic.png" alt="NumPy Contributor comic book cover">
     </a>
 </div>
 
 <style>
-    .comic {
+    .comic img {
         max-width: 20%;
-        max-height: 20%;
```
agriyakhetarpal: We are not changing the image's aspect ratio anywhere, so the max-height rule is unnecessary and has no effect. The max-width rule will handle the height automatically.

Issue #890, opened 2026-01-09
News page readability could be improved with clearer visual separation


hari7niv:
The News page currently presents release notes as a continuous block of text, which makes it hard to scan. I’d like to improve readability by adding lightweight containers or visual separation around each release, without changing content or overall layout. Would this be acceptable?

rgommers:
It has section headers in bold with anchors (https://numpy.org/news/#numpy-1260-released), so it's a normal page rather than "continuous block of text" I'd say. So it doesn't stand out as particularly bad. However, if you can make it look nicer in an easy to review and maintain way, then sure why not:)

PR #891, opened 2026-01-09
Improve News page readability via section separation

hari7niv:
Fixes gh-890

This PR improves the readability of the News page by adding subtle visual separation between individual release sections. The change is CSS-only, scoped to the News page, and does not modify content, templates, or the theme.


melissawm (approved):
I like it! Thanks @hari7niv !

PR #911, opened 2026-09-10
Polish wording in contribute.md

InessaPawson:


Shekharrajak (approved):
LGTM apart from CI checks.

melissawm (approved):
Thank you, @InessaPawson !

melissawm:
CI failure seems to be a timeout - I can't re-trigger but I bet this would be a temporary error.

content/en/contribute.md
```diff
@@ -14,21 +14,23 @@ list](https://mail.python.org/mailman/listinfo/numpy-discussion) or
...
 
-Those are our preferred channels (open source is open by nature), but
-if you prefer to talk privately, contact our community coordinators at
-<numpy-team@googlegroups.com> or on [Slack](https://numpy-team.slack.com)
-(write  <numpy-team@googlegroups.com> for an invite).
+Those are our preferred channels (open source is open by nature). We also
+recognize that reaching out in public can feel a little daunting,
+if you are new to the project or to open source. If you’d feel more comfortable
+starting with a private conversation, you are welcome to reach out to our
+active maintainers via <numpy-team@googlegroups.com> or on
```
seberg: I don't think this is active maintainers, but a rather smaller group?  In that sense I think "community coordinaters" might be better even?

I would suggest putting slack first and maybe mention the maintainers mailing list send? (I'll guess that anyone can send to that.)

seberg:
Thanks Inessa, let's put it in, I like the re-wording!

One thing we may want to follow-up: IIRC very few people get the `numpy-team` mails (in a sense "community managers" actually grasped that better, but is probably not a great fit either).
Maybe that just is: we may want to add people (including with the right to invite to slack) and then _could_ consider moving back to "community managers" here.

seberg:
Not clear why CI is failing to me, so I'll try to close/reopen?  (Sorry I don't do numpy.org stuff often enough)

